PR #31, opened 2014-10-08
Restart estatus on power off

814ckf0x:
Some times when using Caffeine, I realized that every time when I restart the PC and forgot to disable Caffeine, Caffeine starts in enabled mode, and I think this could be dangerous for security, I usually use caffeine just for some applications and just for a session.
So I added a new feature leaving the decision to the end-user whether automatically disable the Caffeine on power-off.
I forgot to say that English is not actually my mother language, so I apologize if I make some mistakes.


eljenso:
This feature would be great to have. I was already looking for a way to reset caffeine on reboot.

Will this PR merged sometime?

814ckf0x:
Eljenso, Thanks for your comment, I think this PR is not going to be merged since I made a lot of mistakes coding it and committing it. no much experience with neither Git nor JavaScript. But I've upgraded my skills, so I promise I'm going to review the code and start a new Pull Request with clean code and best practices.

eonpatapon:
I think it should be the default behaviour. No need to make an option for that I suppose

eljenso:
Created a new issue. I think this PR should be closed.

PR #46, opened 2015-04-19
port lib.js to convenience.js

dffischer:
While [packaging this extension for ArchLinux](https://aur.archlinux.org/packages/gnome-shell-extension-caffeine-git/), I noticed that it does not use the [`convenience.js` taken from GNOME's own extensions](https://github.com/GNOME/gnome-shell-extensions/blob/master/lib/convenience.js) but a very similar `lib.js`.

It only took a few lines porting to `convenience.js`, which allows this project to follow upstream on changes in `convenience.js` and gives packagers the possibility to install the common file only once. It even simplified the code a little by getting rid of some function parameters now stored in `metadata.json`.


l300lvl:
i hadn't noticed before but while looking at this, you are using major version numbers for dev versions, 

i.e if you were to add 3.15 to metadata.json>shell-version then this would never load.

when supporting dev versions you must explicitly use minor versions in the metadata, i.e

3.15 would be 3.15.91, 3.15.92

i know this is unrelated to the pull request but it should be changed as well, if it is to be packaged then it should actually support the versions metadata claims to support, otherwise packagers will be stumped...simply put, you might be better off removing all the dev version support from metadata so it doesnt confuse anyone.

dffischer:
Thanks for the quick processing.

@l300lvl: When packaging, I rewrite the version numbers anyway, because many extensions work with more versions than they claim to.

PR #54, opened 2016-01-27
Added support for gnome-shell-3.18.3-1 on Fedora 23

fmikker:
Added version string to be able to install on F23


eonpatapon:
The "3.18" version should include all releases

Issue #53, opened 2016-01-26
Disable caffeine automatically on manual lock


ijanos:
Every time when I lock my screen it means that I don't need the caffeine feature any more but I tend to forget to turn off caffeine.

It would be great if it automatically disables itself when I lock the screen or put the machine to sleep manually.

eljenso:
Caffeine is now disabled after locking.
Not sure about sleeping, can't test this.

ijanos:
Great! Thank you! I'm waiting for the e.g.o update then I will check this out and close this.

eonpatapon:
Updated on e.g.o

ijanos:
Works great! Thank you!

Tom743:
Can you add an option to make this happen again? I just want to download a big file on my laptop and not use so much battery, would be nice if I could lock the screen (that would turn it off) and keep downloading the file (preventing the computer from suspending).

PR #57, opened 2016-02-17
Remove state restore

### LINKED ISSUE #56: Disable caffeine on shutdown
Currently caffeine restores the state after reboot, which can be confusing.
As mentioned in #31, caffeine should be disabled at some point during a reboot.

eljenso:
With this change, caffeine always starts disabled.
Would resolve #56.


eljenso:
As I just tested, this would also resolve #53.

PR #62, opened 2016-03-16
New update script for locale files

JuniorPolegato:
I completed all needed routines to create a PO template file (.pot), update all PO file for each language and compile to .mo. I also removed .po and .mo files which were wrong in "locale" directory.


eonpatapon:
Nice, thanks

Issue #83, opened 2017-03-13
Preferences dialog won't load on GNOME 3.24


greatquux:
GNOME 3.24 contains a newer version of GJS which has stricter JavaScript rules, e.g. you can't redeclare variables.  When trying to bring up the preferences box I get:

TypeError: redeclaration of let hbox

Stack trace:
  Application<._getExtensionPrefsModule@resource:///org/gnome/shell/extensionPrefs/main.js:74:13
  wrapper@resource:///org/gnome/gjs/modules/lang.js:178:22
  Application<._selectExtension@resource:///org/gnome/shell/extensionPrefs/main.js:89:31
  wrapper@resource:///org/gnome/gjs/modules/lang.js:178:22
  Application<._onCommandLine@resource:///org/gnome/shell/extensionPrefs/main.js:243:17
  wrapper@resource:///org/gnome/gjs/modules/lang.js:178:22
  main@resource:///org/gnome/shell/extensionPrefs/main.js:399:5
  @<main>:1:49

PHLAK:
Same issue here. Waiting for update.

PR #84, opened 2017-04-10
Initial version of code compatible with Gnome Shell 3.24 (mozjs38)

jpraczyk:
This introduces changes required by mozjs38 which appears to be more strict when it comes to syntax rules.

I've also tested the code on Gnome Shell 3.18 and it seems to still be working alright.

One note, I'm not a developer (or a javascript developer) so please, take your time and browse through the changes to see if I didn't mess up something critical here. I've only pieced together the knowledge required to fix the preferences panel for mozjs38.

It seems that [TopIcons Plus had a similar issue](https://github.com/phocean/TopIcons-plus/issues/51) and that allowed me to hack this "fix" together :)

This "fixes" the issue reported in https://github.com/eonpatapon/gnome-shell-extension-caffeine/issues/83

Thanks for developing this awesome extension :+1:

### DIFF
--- a/caffeine@patapon.info/metadata.json
+++ b/caffeine@patapon.info/metadata.json
@@ -1,5 +1,5 @@
 {
-"shell-version": ["3.10", "3.11", "3.12", "3.13", "3.14", "3.16", "3.18", "3.20", "3.22"],
+"shell-version": ["3.10", "3.11", "3.12", "3.13", "3.14", "3.16", "3.18", "3.20", "3.22", "3.24"],
 "uuid": "caffeine@patapon.info",
 "name": "Caffeine",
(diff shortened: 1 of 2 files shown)


jbicha:
I am not the maintainer of this extension, but I did push this patch to Ubuntu 17.04 because this extension is packaged in Debian and Ubuntu and Ubuntu 17.04 is releasing with GNOME 3.24 this week.

jpraczyk:
@jbicha Awesome!

I am Ubuntu GNOME 17.04 user myself so this was the reason I've tried patching the preferences window for this extension.

daniviga:
@jpraczyk thanks!

This code fixes the issue on Fedora 26 alpha too.

Issue #68, opened 2016-06-12
(Option to?) Integrate with system menu


dsboger-zz:
This is a proposal for a possible aesthetic enhancement, so of minor importance. Feel free to "won't fix" it.

Would it be possible to move the enable/disable action into the system menu, similar to e.g. the wifi menu? Also, the indicator would be invisible when caffeine is disabled, but would be visible together with the other system indicators (e.g. wifi, volume, battery, etc.) when active?

Thanks for your efforts!

blahax2g:
It can be something like in this issue: #81 
![caffeine-menu](https://user-images.githubusercontent.com/29534847/35778948-d31c859a-09c5-11e8-9999-63af6655224e.png)

eonpatapon:
We can close this since https://github.com/eonpatapon/gnome-shell-extension-caffeine/pull/216 is merged

PR #108, opened 2017-10-08
Integrate Caffeine with system status menu

dsboger:
This is a working proposal for #68. What do you think?

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -27,10 +27,12 @@ const Gio = imports.gi.Gio;
 const Main = imports.ui.main;
 const Mainloop = imports.mainloop;
 const PanelMenu = imports.ui.panelMenu;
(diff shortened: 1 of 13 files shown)


dffischer:
I tested this on my system and and like it quite much. As I use a whole bunch of extensions, this is a welcome option to clean up my bar.

It should be optional, though, so that users may keep the icon variant if they like it. There already is an option in the schema and the settings dialog to show the bar icon. Maybe you could copy that for the menu entry, so that both can be activated separately or together.

dsboger:
@dffischer This was just a proof-of-concept patch and I have no intention of working on this any further. Owner is not responding for months, so I think this extension is abandoned and I don't want to fork it. But thanks for testing!

eonpatapon:
We can close this since https://github.com/eonpatapon/gnome-shell-extension-caffeine/pull/216 is merged

PR #135, opened 2019-03-17
Add support for Gnome 3.32

julianxhokaxhiu:
Tested, works.

### DIFF
--- a/caffeine@patapon.info/metadata.json
+++ b/caffeine@patapon.info/metadata.json
@@ -12,7 +12,8 @@
     "3.24",
     "3.26",
     "3.28",
-    "3.30"
+    "3.30",
+    "3.32"
   ],
   "uuid": "caffeine@patapon.info",
   "name": "Caffeine",


aifrs12:
Fixed the no icon bug?

julianxhokaxhiu:
On my side all the icons works just fine. Although I'm using another two extensions to add support for systray icons ( appindicator and TopIcons Plus ), so it could be I'm not impacted by this.

Nevertheless everything works smooth on 3.32.

Mr-Monkey:
This change doesn't seem to fix the missing icon for me. Anyone else?

aifrs12:
Still not working. The Extension works great, but no icon on top bar

julianxhokaxhiu:
Again, the icon will never appear as it requires a systray, which GNOME definitely removed some releases ago.

If you want the icon, you need other two extensions which re-create the systray support ( appindicator and/or TopIcons Plus ).

I don't see how this PR should be kept open because of an Icon support. This wasn't "officially" supported also on 3.30, yet the extension was tagged.

NickCis:
#137 Fixes the icon

eonpatapon:
Fixed by #137

PR #139, opened 2019-04-02
Update JS syntax

NickCis:
It updates JS syntax to met [new standards](https://2017.guadec.org/wp-content/uploads/2017/06/Philip-Chimento-GUADEC-2017-Modern-Javascript-in-GNOME.pdf)

- Removes the use of `Lang.bind` (uses native `.bind`)
- Uses arrow functions instead of binding anonymous functions
- Uses classes instead of `Lang.Class`
- Uses `.includes( ... )` instead of `.indexOf( ... ) !== -1`
- Uses string templates

PR #144, opened 2019-04-09
Fix `shell-version` data in `metadata.json`

NickCis:
As @jagaudin and @mirko-laruina have pointed out in #139 , `GObject.registerClass` is available since `1.49.90` (gjs version). The first gnome-shell which requires an equal or higher version is `3.31.92`.

This PR erases not supported gnome-shell versions.

Sorry for the inconveniences that caused by the refactor!


jagaudin:
@NickCis good job

PR #152, opened 2020-02-25
Update Brazilian Portuguese translation

rffontenelle:


eonpatapon:
Can you also run the `./update-locale.sh` script ?

Thanks

rffontenelle:
I ran it just now and nothing changed in any .po file, possible because you already did it 3 days ago. Not even the POT-Creating-Date go updated, probably because of the `msgmerge`'s `-U` flag.

eonpatapon:
Right, I thought that the .mo files were commited in the repo also, but it turns out no :)

Thanks!

PR #179, opened 2021-04-10
Gtk4 / GNOME 40 Support

kaechele:
Reworked this again.

This time using `eslint` to check for code smells and also moved away from the ui files again to keep things simpler (given that Glade does not work with Gtk4 as of writing) and in one place.
Tested all functionality and it is working for me.
Tested it on GNOME 3.38 and the prefs panel no longer functions. So it seems there needs to be some compat code weaved into the `prefs.js` to make it work with older versions of GNOME Shell.

If the requirement on GNOME 40 is acceptable feel free to merge, otherwise I may need to see when I find more time to look at the compat stuff needed.


caffeine@patapon.info/schemas/org.gnome.shell.extensions.caffeine.gschema.xml
```diff
@@ -31,15 +37,10 @@
         <summary>Enable when a fullscreen application is running</summary>
         <description>Enable when a fullscreen application is running</description>
     </key>
-    <key type="b" name="control-nightlight">
-        <default>false</default>
+    <key name="control-nightlight" enum="org.gnome.shell.extensions.caffeine.control-nightlight">
+        <default>"never"</default>
```
eonpatapon, line 41: Wouldn't this be a problem while upgrading ? Ie old setting type won't match new setting type

eonpatapon (commented):
Thanks for working on this, it looks great, one question though

caffeine@patapon.info/schemas/org.gnome.shell.extensions.caffeine.gschema.xml
kaechele: OK, so what happens if the value is already set with the wrong type is that it will be deleted and recreated using the correct type. This means the user's setting is lost.
I can look to rewrite this particular part to convert the user's current setting.
I will most likely need to rename the new setting to something else because reassigning the type will lead to the aforementioned behaviour upon loading the modified schema.

kaechele: I've addressed this in the latest commit.

eonpatapon: Thanks a lot!

PR #183, opened 2021-06-02
Changing the shell version

Skaldebane:
There's no need to add the subversion here, and using a plain "40" supports all versions of GNOME 40.


eonpatapon:
Was done by another PR

PR #193, opened 2022-03-19
add gnome 42 support

TheTakylo:


stuarthayhurst:
This is in `master` now via another PR (#194), so can probably be closed

PR #196, opened 2022-03-31
Update README for version and build instructions

stuarthayhurst:
- As GNOME 42 is supported by the extension, the README has been updated to actually list support for it
 - A set of instructions to build and install the extension using `make` has also been added, for simplicity

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@
 
 Fill the cup to inhibit auto suspend and screensaver.
 
-This extension supports gnome-shell 3.4 to 41
+This extension supports gnome-shell 3.4 to 42
 
-* master: 40 -> 41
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,7 +24,11 @@ screensaver off.
 https://extensions.gnome.org/extension/517/caffeine/
 
 ## Installation from git
-
+```
+make build
+make install
+```
+Alternatively:
```
eonpatapon, line 31: I think we can remove the manual commands below and rely on the Makefile only

stuarthayhurst: Done :)

eonpatapon: Thanks!

Feel free to open an other PR if you'd like to improve the README more

PR #207, opened 2022-06-22
Add terminal support

### LINKED ISSUE #148: termainal support
is caffine able to be toggled from a terminal command?

### LINKED ISSUE #189: Let user define a shortkey which trigger Caffine. Save a LOT of time.
I love this extension so far. It's simple and perfect. 
But there is something annoying when u have to click Caffeine at least 20 times a day.
20 x 4 sec = 80 sec = 1min + 20sec. For vim users, this is a lot. Anyone who is more keyboard-driven, this can be unacceptable. 
I think it would be really cool if there is some sort of command to activate Caffeine mode.
So I can use this "command" in `GNOME settings > Keyboard > View and Customize Shortcuts > Customize Shortcuts > +` to add our Caffeine command to activate and deactivate extension. Sort of like:
```
gnome-extensions activate caffeine@patapon.info
gnome-extensions deactivate caffeine@patapon.info
```

Another way to roll this feature is, to add another option here in this section:
image
I have seen other gnome extensions successfully rolling their own shortkey (like [Dash-To-Panel](https://github.com/home-sweet-gnome/dash-to-panel): User can hide/unhide panel using their own shortkey).

Something like:
Super + C = Activate Caffine
Super + C = Deactive Caffine         -----------  _(if caffine is already activate)_

notabugwontfix:
Had so much fun with states.

- Fixes #148
- Fixes #189
- Get current state: `gsettings --schemadir ~/.local/share/gnome-shell/extensions/caffeine@patapon.info/schemas/ get org.gnome.shell.extensions.caffeine user-enabled`
- Enable: `gsettings --schemadir ~/.local/share/gnome-shell/extensions/caffeine@patapon.info/schemas/ set org.gnome.shell.extensions.caffeine user-enabled true`
- Disable: `gsettings --schemadir ~/.local/share/gnome-shell/extensions/caffeine@patapon.info/schemas/ set org.gnome.shell.extensions.caffeine user-enabled false`

`--schemadir` must be a path to the extension `schemas` directory. It may be different on your system.

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -174,7 +174,7 @@ class Caffeine extends PanelMenu.Button {
         this._appData = new Map();
 
         this._settings.connect(`changed::${INHIBIT_APPS_KEY}`, this._updateAppConfigs.bind(this));
-        this._settings.connect(`changed::${USER_ENABLED_KEY}`, this._updateState.bind(this));
+        this._settings.connect(`changed::${USER_ENABLED_KEY}`, this._updateUserState.bind(this));
 
         this._updateAppConfigs();
     }
@@ -320,7 +320,7 @@ class Caffeine extends PanelMenu.Button {
         this._updateAppData();
     }
 
-    _updateState() {
+    _updateUserState() {
         if (this._settings.get_boolean(USER_ENABLED_KEY) !== this._userState) {
             this._userState = !this._userState;
             this.toggleState();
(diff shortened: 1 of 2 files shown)


caffeine@patapon.info/extension.js
```diff
@@ -317,6 +320,18 @@ class Caffeine extends PanelMenu.Button {
         this._updateAppData();
     }
 
+    _updateState() {
```
eonpatapon, line 323:
```suggestion
    _updateUserState() {
```

eonpatapon (commented):
LGTM, thanks

PR #216, opened 2022-10-15
Migrate to the new Quick settings menu and more

pakaoraki:
Gnome 43 introduced a new quick settings toggle menu and I propose to rewrite the extension to follow this evolution.

screenshot

Changelog:
-Drop Panel menu and migrate to QuickSettingsMenu.
-Scroll up/down the indicator icon to enable/disable caffeine instead of clicking it (click open the toggle menu now).
-Update 'Show status icon' option with one new state 'only when active'.
-Upgrade notifications to the latest Gnome OSD notification.
-Update README and screenshots.

Pros/cons: using a toggle menu instead of panel icon cost one more click, but it's still possible to scroll directly on the status icon so it still quickly accessible.

My repo: [https://github.com/pakaoraki/gnome-shell-extension-caffeine](https://github.com/pakaoraki/gnome-shell-extension-caffeine)

I use this extension since a while now, thanks for everybody who works on it. Feel free to merge or cherry-pick any code that you want to use in Caffeine. I'm please for any feedback or modification on this pull request.

Cheers

PR #220, opened 2023-01-12
OSD notification updates

### LINKED ISSUE #173: Change wording of notification to "Caffeine enabled", not "Auto suspend and screensaver disabled"
When you activate this program called Caffeine, you get this notification:

> Auto suspend and screensaver disabled

This is confusing to me. Didn't I just activate something? Why is it saying that something is "disabled"?

When you turn off Caffeine functionality, you get this notification:

> Auto suspend and screensaver enabled

Again this is confusing to me.

The icon is also confusing to me. Which one is "on"? Is it the icon of the full cup? Or the icon of the cup with steam but not liquid?

Here is my proposal:

1. Change the text of the notifications to say:

> Caffeine is now preventing your computer from automatically suspending or launching a screensaver

> Caffeine is no longer preventing your computer from automatically suspending or launching a screensaver

2. Regarding the icons, I would make the one that represents caffeine being off (normal suspend behaviour) to be an empty cup, and the one representing on to be the icon with steam.

stuarthayhurst:
This changes "Auto suspend and screensaver disabled" to "Caffeine enabled", as having disabled in the notification is confusing when something was enabled. (Closes #173)
This'll need the locales updating again, too

Also, what about this?
```
                    // Force notification here if disable in prefs
                    if (!this._settings.get_boolean(SHOW_NOTIFICATIONS_KEY)) {
                        this._sendOSDNotification(true);
                    }
```
Personally, I find it a bit annoying that it sends a notification when scrolling on the icon, even though it's disabled. If the user has disabled the notifications, why would they then want them specifically for scrolling? I'll happily add this to the PR if you agree :)

PS: Did you see my comment in #216? I think the `gnome-shell-40-43` branch needs creating from commit `86116f3f7d2c7959abad4b433db9ff9844d93d70`, unless you were planning to do this on release.


README.md
```diff
@@ -43,10 +43,10 @@ make install
 ### <ins>Scroll Indicator Icon</ins>
 ![Screenshot](screenshots/screenshot-scroll-up.png)![Screenshot](screenshots/screenshot-scroll-down.png)
 
-You can scroll on the indicator icon to Enable/disable auto suspend:
+You can scroll on the indicator icon to enable / disable auto suspend:
 
-- Scroll UP -> Filled cup : auto suspend and screensaver off. 
-- Scroll DOWN -> Empty cup : normal auto suspend and screensaver.
+- Scroll up -> Filled cup: Enable Caffeine
+- Scroll down -> Empty cup: Disable Caffeine
```
eonpatapon, line 49:
```suggestion
- Scroll up -> Filled cup: Caffeine enabled
- Scroll down -> Empty cup: Caffeine disabled
```

eonpatapon:
> Personally, I find it a bit annoying that it sends a notification when scrolling on the icon, even though it's disabled. If the user has disabled the notifications, why would they then want them specifically for scrolling? I'll happily add this to the PR if you agree :)

:+1:

pakaoraki:
It was just a personal choice: I did this because when you scroll on volume icon, it show OSD notification. So to keep the same coherent behavior in Quick Setting menu, I choose force notification for scrolling on Caffeine icon. But, in the other hand, when notification are disabled, it suppose to not have notification at all, so I think we can definitely remove that.

README.md
stuarthayhurst: Fixed, thanks :)

stuarthayhurst:
That should be everything now :)

PR #223, opened 2023-02-01
Disconnect from timeouts in destroy()

stuarthayhurst:
As discussed in #216, this should fix the issues for release


eonpatapon:
Looks good, thanks!

PR #228, opened 2023-02-08
Added more translations for pt_BR

Braayy:
Hi! I had some spare time, so i've added more translations on pt_BR for the project.


eonpatapon:
Thanks a lot!

PR #229, opened 2023-02-10
Use 15, 30 and 60 minute timers (closes #224)

### LINKED ISSUE #224: Times adjustable
Hello,

I love this app well done and thank you.  I have one small nit pick, I really don't care for applications that choose defaults such as times of 5, 10 and 30 minutes without providing a way to adjust them.  Would you please consider adding this.

I find the defaults of no use to me, 30, 60 an 120 would be more usable for what I do.  Being able to adjust our own times would be a desirable feature.

Thank you

stuarthayhurst:
Implements the suggestion from #224 

If this needs the locales to be regenerated, LMK and I can do that if you want :)

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -119,9 +119,9 @@ const AppsTrigger = {
 };
 
 const TIMERS = [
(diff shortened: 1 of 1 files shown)


stuarthayhurst:
Seems this is handled by #230, feel free to close

PR #231, opened 2023-02-18
GNOME 44

DaPigGuy:


stuarthayhurst:
Can you do this with a condition? It seems a bit soon to drop GNOME 43 support, when the changes are small and both could be supported.

stuarthayhurst:
You could use something like `const ShellVersion = parseFloat(imports.misc.config.PACKAGE_VERSION);` at the start to get the shell version, then do something similar to https://github.com/stuarthayhurst/privacy-menu-extension/commit/3aa41d49222e5794c90dd33bf6fb4afdfca79c86 to add the condition for both GNOME 43 and 44 support.

Keeping GNOME 43 support is probably a good idea as Debian 12 is using GNOME 43, and not much work is required to keep both.

DaPigGuy:
> Can you do this with a condition? It seems a bit soon to drop GNOME 43 support, when the changes are small and both could be supported.

Done. The `clicked` handler doesn't require a condition as it appears to have no side effects on GNOME 43. On GNOME 44, however, clicking the toggle has no effect without it, but I was unable to identify what changed to cause that.

DaPigGuy:
The Caffeine quick settings toggle is appended after the Background Apps menu button which is hidden by default until the app count is >1. Quick settings provided by extensions look quite awkward when that is visible, but I'm not entirely sure how to handle that.

stuarthayhurst:
> Something I noticed, the Caffeine quick settings toggle is appended after the Background Apps menu button which is hidden by default until the app count is >1. Quick settings provided by extensions look quite awkward when that is visible, but I'm not entirely sure how to handle that, and if that should even be handled on an individual basis by extensions...

Seems a bit impractical to fix by every extension, that should probably be reported upstream.

Since this works for GNOME 43 and 44, I'm happy with this. Maybe @pakaoraki would know more about the `clicked` handler?

DaPigGuy:
One more thing from me, the quick setting pills now support a subtitle,  should we maybe use that when Caffeine is toggled?

stuarthayhurst:
Might be a good idea, it would probably fit in better with the rest of the shell. I haven't played around with the testing version much, but if the other shell quick settings have a subtitle when they're toggled, it's probably worth adding.

stuarthayhurst:
Actually, that might be better left to another PR, as the time remaining on any set timer could be displayed there instead?

pakaoraki:
Hi,

Thanks for this update.

I reckon it would be nice to have a separate branch for testing gnome44, at least until the official release in march: then we can merge. 

About the `clicked` handler, it could be safer to wait updated official documentation [here](https://gjs.guide/extensions/) as Gnome will publish an extension migration guide like they did for gnome 43 ([see this](https://gjs.guide/extensions/upgrading/gnome-shell-43.html#quick-settings)). This behavior can still change maybe, not sure. 
Meanwhile, I'll suggest to replace  this : 

```
this.connect('clicked', () =>     
    this._settings.set_boolean(TOGGLE_STATE_KEY, !this.checked);
});
```
with
```
this.connect('clicked', () => {
    this.checked = !this.checked;
});
```

The TOGGLE_STATE_KEY boolean is still handled by:
```
this._settings.bind(`${TOGGLE_STATE_KEY}`,
            this, 'checked',
            Gio.SettingsBindFlags.DEFAULT);
```

About the new subtitle feature, I think it could be great to use it too.

> but if the other shell quick settings have a subtitle when they're toggled, it's probably worth adding

I agree, it could also be used for countdown when timer is on for example.

stuarthayhurst:
After testing, the 'Dark Mode' toggle (others too, this was just the first one that's purely an on / off toggle that I noticed) doesn't use the subtitle for enabled, so it's probably best to use it only for a timer countdown, to stay consistent with the rest of the shell.

caffeine@patapon.info/extension.js
```diff
@@ -147,7 +148,10 @@ const CaffeineToggle = GObject.registerClass(
 class CaffeineToggle extends QuickSettings.QuickMenuToggle {
     _init() {
         super._init({
-            label: IndicatorName,
+            // The 'label' property was renamed to 'title' in GNOME 44 but quick settings have otherwise 
+            // not been changed. The below line allows support for both GNOME 43 and 44+ by using the 
+            // appropriate property name based on the GNOME version.
+            [Number(Config.PACKAGE_VERSION.split('.')[0]) >= 44 ? 'title' : 'label']: IndicatorName,
```
petar-v, line 154: Now there is also a `subtitle` property. This could be used to display the countdown timer. ;)

petar-v: Also, thanks for updating that to 44 so quickly! :D

petar-v:
> You could use something like `const ShellVersion = parseFloat(imports.misc.config.PACKAGE_VERSION);` at the start to get the shell version, then do something similar to [stuarthayhurst/privacy-menu-extension@3aa41d4](https://github.com/stuarthayhurst/privacy-menu-extension/commit/3aa41d49222e5794c90dd33bf6fb4afdfca79c86) to add the condition for both GNOME 43 and 44 support.
> 
> Keeping GNOME 43 support is probably a good idea as Debian 12 is using GNOME 43, and not much work is required to keep both.

By the way, for the `_init({...})` parameter, you can pass both `label` and `title` to maintain backward compatibility since it's not a typed js object. Depending on the gnome version API, it will pick up whichever property it wants. Still, somewhat clearer to use a version check. 

Alternatively, there can be a branch for gnome 43 and another one for gnome 44 with the respective versions in the metadata file.

DaPigGuy:
https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/2664 Looks like this should resolve the issue bind issue on GNOME 44. I'll keep the workaround for now, and we can perhaps merge this PR after the 44-rc release.

yochananmarqos:
With or without this PR applied using GNOME 44.rc, it still complains:

```
theme.set_custom_theme is not a function
```

DaPigGuy:
I unfortunately am not able to build gnome-shell 44.rc at the moment. :( If anyone wants to have a go at that issue, feel free.

stuarthayhurst:
Works with `theme.set_theme_name` instead, as of GNOME 44.rc

DaPigGuy:
Hm, probably a result of gnome-shell switching away from GTK3. Thanks for that, latest commit should work on `44.rc` :smile:

stuarthayhurst:
Broken on GNOME 43 now, I guess we'll need a condition for that call to `theme.set_theme_name`

DaPigGuy:
Yep, realized after pushing, fixed :)

stuarthayhurst:
Erm, that's a new error:
```
Mar 08 18:50:14 snakeroot gnome-shell[23260]: JS ERROR: TypeError: Object 0x2c300473d1c0 is not a subclass of GObject_Object, it's a GjsGlobal
                                              Gtk.IconTheme.get_default@/home/ratus5/.local/share/gnome-shell/extensions/caffeine@patapon.info/extension.js:147:53
                                              getDefaultTheme@/home/ratus5/.local/share/gnome-shell/extensions/appindicatorsupport@rgcjonas.gmail.com/util.js:269:44
                                              _init@/home/ratus5/.local/share/gnome-shell/extensions/appindicatorsupport@rgcjonas.gmail.com/appIndicator.js:979:32
                                              AppIndicatorsIconActor@/home/ratus5/.local/share/gnome-shell/extensions/appindicatorsupport@rgcjonas.gmail.com/appIndicator.js:926:1
                                              _init@/home/ratus5/.local/share/gnome-shell/extensions/appindicatorsupport@rgcjonas.gmail.com/indicatorStatusIcon.js:233:13
                                              ButtonBox@resource:///org/gnome/shell/ui/panelMenu.js:11:1
                                              PanelMenuButton@resource:///org/gnome/shell/ui/panelMenu.js:97:4
                                              AppIndicatorsIndicatorBaseStatusIcon@/home/ratus5/.local/share/gnome-shell/extensions/appindicatorsupport@rgcjonas.gmail.com/indicatorStatusIcon.js:72:1
                                              AppIndicatorsIndicatorStatusIcon@/home/ratus5/.local/share/gnome-shell/extensions/appindicatorsupport@rgcjonas.gmail.com/indicatorStatusIcon.js:230:1
                                              _registerItem@/home/ratus5/.local/share/gnome-shell/extensions/appindicatorsupport@rgcjonas.gmail.com/statusNotifierWatcher.js:117:32
                                              async*_ensureItemRegistered@/home/ratus5/.local/share/gnome-shell/extensions/appindicatorsupport@rgcjonas.gmail.com/statusNotifierWatcher.js:142:20
                                              RegisterStatusNotifierItemAsync@/home/ratus5/.local/share/gnome-shell/extensions/appindicatorsupport@rgcjonas.gmail.com/statusNotifierWatcher.js:207:24
                                              _handleMethodCall@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:368:35
                                              _wrapJSObject/<@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:401:34
                                              Caused by: Error: This JS object wrapper isn't wrapping a GObject. If this is a custom subclass, are you sure you chained up to the parent _init properly?
```

yochananmarqos:
Yep, same error here on 44.rc

```
Object 0x2987cd63d1c0 is not a subclass of GObject_Object, it's a GjsGlobal
```

DaPigGuy:
Caused by the null-coalescing operator for whatever reason

stuarthayhurst:
All working on GNOME 43 for me

yochananmarqos:
It's working on GNOME 44.rc as well. :+1:

stuarthayhurst:
Since GNOME 44.rc is out, if it's still working on GNOME 44.rc, is this ready to merge, and add the labels later?

DaPigGuy:
Yes, ready to merge

stuarthayhurst:
Cool, @eonpatapon feel free to review this whenever you get time :)

eonpatapon (approved):
LGTM

PR #241, opened 2023-03-12
Port build system to `gnome-extensions` tool, setup pipeline

stuarthayhurst:
- Added `lint-fix` to phony make targets
 - Added 4 workflows: (runs on commit and pull request)
   - Shellcheck - runs shellcheck on any bash scripts
   - Build system test - runs all build system targets, except `lint` and `lint-fix`
   - Build and upload extension - builds extension bundle, then uploads as an artifact (in the Actions tab), useful for giving test build to users
   - Linter - runs eslint on the codebase, from the build system target

The linter target is currently failing, with ~340 issues. I had a quick scan of them, and some are valid, but some I don't completely agree with. The linter config is set to fail when 'unnecessary' curly braces are added (like in 1 line if statement). Personally, I think the curly braces should be there for clarity and consistency, but that's not the point of this PR.

Let me know you you guys think of this :)


stuarthayhurst:
Seems it doesn't run the workflows on the PR they're introduce on, you can see the results on my personal fork, [here](https://github.com/stuarthayhurst/gnome-shell-extension-caffeine/actions)

stuarthayhurst:
After this PR is sorted, I'll start a draft PR to begin fixing the issues the linter has with the code, and sort the linting rules, if required

.github/workflows/test-build-system.yml
```diff
@@ -0,0 +1,30 @@
+# This workflow will thoroughly test the build system
+name: Complete build system test
```
eonpatapon, line 2: Seems a bit overkill but why not...

.github/workflows/test-build-system.yml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install build dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install --no-install-recommends make gettext gnome-shell libglib2.0-bin
```
eonpatapon, line 18: Why do you need `gnome-shell` ?

stuarthayhurst: Good catch, we don't. This was ported from my extensions, where I used `gnome-extensions pack` and `gnome-extensions install` for the `package` and `install` targets. It might be worth porting to `gnome-extensions` at some point, as that might catch issues like #244 before release, but I can do that in another PR if you'd rather.

.github/workflows/test-build-system.yml
stuarthayhurst: Ported from another project, I guess it's there for any other targets added in future?

.github/workflows/test-build-system.yml
eonpatapon: If `gnome-extensions` provides more guarantees why not. What's bothering me is downloading `gnome-shell` + all dependencies for a simple check...

stuarthayhurst: The `--no-install-recommends` should help mitigate the number of dependencies. This workflow could be grouped with the test build & upload, to avoid setting up the same dependencies twice.

.github/workflows/build-extension.yml
```diff
@@ -0,0 +1,27 @@
+# This workflow will build the extension and upload an artifact
+name: Build and upload extension
+
+on:
+  push:
+    branches: '**'
+  pull_request:
+    branches: '**'
```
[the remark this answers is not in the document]
stuarthayhurst: I'll go with running the workflows on PRs for now, let me know if you'd rather something else, and if that's specific to any of the workflows

stuarthayhurst:
Well it would seem I suck at `git`, oops.

I've grouped the build system test build and the extension upload, as well as limiting the workflows to running on PRs.
The build system now uses `gnome-extensions`, which took a slight restructure of `locale`. It does mean that `update-locale.sh` doesn't need to create the `.mo` files anymore, and the file structure was be changed to have all `.po` files named as their language, in 1 directory. Produces an identical extension bundle when tested, and should hopefully mean that issues like #244 get caught during testing.

.github/workflows/build-extension.yml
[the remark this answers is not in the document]
eonpatapon: Fine with me

.github/workflows/test-build-system.yml
eonpatapon: Still there's a lot of dependencies and it's a waste of resources to download them over and over again on each run. At least we should setup some cache for apt packages (ex: https://github.com/awalsh128/cache-apt-pkgs-action)

If the behaviour of `gnome-extensions` changes from one gnome-shell release of another (not yet officially released) I doubt we will catch it as we will still use the older version. There's probably no good solution here.

Anyway I think it's a good idea to standardize the build/install with the tool provided by gnome-shell.

eonpatapon:
> Currently, the install command would be `make dist` and `make install`. The locales don't need to be updated every build anymore, so perhaps the `glib-compile-schemas` command could be moved into `dist`, and then `dist` renamed to build? That way the build and install instructions stay the same, and `dist` can be replaced with `build`, as the bundle is created for every install now. Let me know what you think of this, I'll tweak the system as required.

I'm fine with that

.github/workflows/test-build-system.yml
stuarthayhurst: Ok, I'll look into the caching the dependencies. True, the workflow probably won't catch issues due to changes in new versions, but manual testing would catch this on the target version, as long as `gnome-extensions`. is used.

.github/workflows/build-extension.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    - name: Build the extension bundle
+      run: |
+        make dist
+
+    - name: Test extension installs
+      run: |
+        make install
+
+    - uses: actions/upload-artifact@v3
```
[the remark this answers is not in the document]
stuarthayhurst: Mm, it's easy enough to build, but this saves users cloning the right fork, switching to the right branch, downloading the dependencies and then building. Building the extensions automatically guarantees that they've got the correct build of the extension, and can install it like any regular GNOME extensions, without worrying about which target will do it.

I can drop this if you'd rather, but since the extension only takes a few KB, and it's built and stored within GitHub infrastructure, I wasn't too worried about unnecessary uploads. AFAIK, it also clears old builds after enough time / rebuilds, and this limit can be configured down.

stuarthayhurst:
With those 2 commits, `make build` and `make install` build and install the extension, so no changes to the regular install workflow. The output of `make build` is the output of the old `make dist`, but using `gnome-extensions`. If translations are changed, `./update-locale.sh` needs to be run manually, unless you want a target added for this.

`make clean` just removes the built extension and compiled gschemas settings

stuarthayhurst:
Well that might work? Testing these things is nearly impossible, but that follows the project's README, and it ran that step in testing, so hopefully it works.

stuarthayhurst:
Rebased on master to fix conflicts, let me know what you want to do about uploading builds and whether you're happy with the dependency caching attempt, then after that I don't have anything else to add :)

stuarthayhurst:
@eonpatapon Soft ping to review this when you get time :)

pakaoraki:
I think it's a really good idea and it will make a great addition to this project. I didn't participate to this PR earlier because I'm not very familiar with Github workflow as you and @eonpatapon seems to know lot better than me on this. 

I clone your repo branch [setup-pipeline](https://github.com/stuarthayhurst/gnome-shell-extension-caffeine/tree/setup-pipeline) for testing: `spellcheck` and `test build` ran fine but I got this error with `lint`.

```
Run make lint
eslint --resolve-plugins-relative-to "/usr/local/lib/node_modules" "caffeine@patapon.info"

Oops! Something went wrong! :(

ESLint: 8.38.0

ESLint couldn't find a configuration file. To set up a configuration file for this project, please run:

    npm init @eslint/config

ESLint looked for configuration files in /home/runner/work/test-gnome-shell/test-gnome-shell/caffeine@patapon.info and its ancestors. If it found none, it then looked in your home directory.

If you think you already have a configuration file or if you need more help, please stop by the ESLint Discord server: https://eslint.org/chat

make: *** [Makefile:25: lint] Error 2
Error: Process completed with exit code 2.
```

It's looking a for a configuration file, I should miss something.

stuarthayhurst:
Where did you get `eslint` from? Using the Debian `eslint` has issues for me, but `sudo npm install -g eslint eslint-plugin-jsdoc` seems to work.

stuarthayhurst:
Fixed my install, running the same version of you now, should be fixed by the latest commit

pakaoraki:
I made a new test today with your last commit, and it ran as expected this time. I forgot to upload .eslintrc.yml in my first attempt.

> Where did you get eslint from? Using the Debian eslint has issues for me, but sudo npm install -g eslint eslint-plugin-jsdoc seems to work.

I simply pushed the new files on my test repo on Github and I didn't change anything to the config. Then I added some PR to test the process.

This is looking good to me. I let @eonpatapon the final review.

stuarthayhurst:
Rebased with the cleanups to the translation files

Makefile
```diff
@@ -1,21 +1,28 @@
+BUNDLE_PATH = "caffeine@patapon.info.zip"
+EXTENSION_DIR = "caffeine@patapon.info"
+
 all: build install
 
-.PHONY: build install lint dist
+.PHONY: build install lint lint-fix
 
 build:
-	./update-locale.sh
```
eonpatapon, line 6: Shouldn't we need to keep this ? Not so sure...

eonpatapon: Ah I guess `gnome-extensions pack` does this!

.github/workflows/linter.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v3
+    - uses: awalsh128/cache-apt-pkgs-action@latest
+      with:
+        packages: make npm
+        version: 1.0
+    - name: Install eslint
+      run: |
+        sudo npm install -g eslint eslint-plugin-jsdoc
```
eonpatapon, line 19: Isn't it working without sudo ?

```suggestion
        npm install -g eslint eslint-plugin-jsdoc
```

eonpatapon:
So all of that is nice, but how are we going to proceed with linting errors ?

We need to make linting work before merging this, what's your take on this ?

Makefile
stuarthayhurst: Yep, `gnome-extensions pack` generates the .mo files, and puts them directly into the bundle. This means we don't need the whole `[LANGAUGE_CODE]/LC_MESSAGES/gnome-shell-extension-caffeine.po`, as it generates them straight from `[LANGUAGE_CODE].po`.

Since we don't need to generate the files manually to make the bundle, we don't need to call `./update-locale.sh` every build, so `make build` won't change the translation files, it'll leave the source alone and just build the extension.

.github/workflows/linter.yml
stuarthayhurst: Ah, that command failed without sudo on my machine, I didn't test within the workflow

stuarthayhurst:
Of course, I didn't want to keep flooding this PR with small JS linting changes, until the main bits were settled. They're mostly many instances of the same error, so it shouldn't take too long to tweak some of the rules and fix any issues.

Should I work on that in this PR? Creating another one for the JS stuff would be theoretically be cleaner, but there are changes to the rules and code in this one already.

PS: The commit I marked that needed to be reverted has been reverted, the PR just doesn't show that unless you're in the commits tab. The workflow only runs on PRs, so I just added a rule to run on each push to test the sudo change, then reverted.

stuarthayhurst:
Also, if you'd rather continue to use `make` targets to do the translation management, I can add a simple `make translations` target to avoid calling the script directly, when translation files need refreshing.

stuarthayhurst:
That should fix ~250 warnings, it's down to 28 now. Most of the remaining issues are just about marking exported objects to stop eslint whining.

stuarthayhurst:
Sorry for creating so much noise with this PR, but I believe that's everything now.

Changes:
 - Added `translations` Makefile target, for completion
 - Added `clean` target to phony targets list, as it was missing
 - Fixed all remaining ESLint issues

eonpatapon:
> Also, if you'd rather continue to use `make` targets to do the translation management, I can add a simple `make translations` target to avoid calling the script directly, when translation files need refreshing.

I think it would be useful to have as we want to update po files before making a release.

update-locale.sh
```diff
@@ -9,11 +9,9 @@ xgettext -j ./*.js -o "$pot" --from-code UTF-8 --no-wrap
 xgettext -j preferences/*.js -o "$pot" --from-code UTF-8 --no-wrap
 xgettext -j schemas/*.xml -o "$pot" --from-code UTF-8 --no-wrap
 
-for locale_lang in locale/*; do
-    po="$locale_lang/LC_MESSAGES/gnome-shell-extension-caffeine.po"
+for po in locale/*; do
     echo "$po"
     msgmerge --backup=off -U "$po" "$pot"
-    msgfmt "$po" -o "${po%po}mo"
```
eonpatapon, line 16: Shouldn't we remove mo files from locale dir as well ?

I see the po files are `locale/lang.po` instead of `locale/lang/LC_MESSAGES/gnome-shell-extension-caffeine.po`, that's something required by gnome-extensions pack?

stuarthayhurst: > I see the po files are locale/lang.po instead of locale/lang/LC_MESSAGES/gnome-shell-extension-caffeine.po, that's something required by gnome-extensions pack?

As far as I know, yes. That's what I've done for my other extensions, and that's been working. I'm assuming the old structure was required because the extension was being zipped manually, but since `gnome-extensions pack` handles `.mo` creation, it picks its own structure.

> Shouldn't we remove mo files from locale dir as well ?

I'm not sure what you mean, I can't see any `.mo` files in `locale/`?

eonpatapon: Sorry there no `mo` files, they are ignored by git

eonpatapon (approved):

Issue #234, opened 2023-03-01
Clicking "infinite" both changes a preference and activates Caffeine mode


Flimm:
**Edit:** Old title of this issue: "infinite" is checked even when Caffeine is disabled

I'm using Caffeine version 44. Have a look at this screenshot from the README:

![screenshot-timer-off](https://user-images.githubusercontent.com/355454/222148720-200f29d3-134c-4ff0-a10f-10c2df9678b1.png)

The timer is off. However, the item "infinite" is checked with a tick symbol ✔️. I don't think any of the items (5:00, 10:00, 30:00 or infinite) should be checked when caffeine is off.

When caffeine is on, the "Caffeine" widget lights up, like here:

![screenshot-timer-on](https://user-images.githubusercontent.com/355454/222148616-d84cf6be-ae3e-47c9-bccd-6444c835b86c.png)

Only when Caffeine is on and the "Caffeine" widget lights up do I expect a tick symbol ✔️ to be used.

Thank you to everyone who has a look at this!

stuarthayhurst:
I agree with eon here, as when Caffeine is inactive, having a tick next to the infinite timer indicates that's the option to be used when enabled. The tick is to indicate the timer preference, where as the blue indicates whether the timer is running (as well as the countdown).

Flimm:
Reading your comments made me go "aha!". I didn't realise other people were viewing the options 5, 10, 30, or infinite in this way. I now have a new suggestion for a design which should satisfy that group of people as well. But first, let me describe the current design.

*Current design (design 1)*:

- Clicking on an option 5, 10, 30 or infinite sets the preference to that time limit and activates caffeine, while also closing the menu. If caffeine is already active, clicking on this option resets the time limit to that option.
- Clicking the main menu item enables caffeine mode using the previously selected time limit, or disables it if it is already active
- The checkmark indicates which preference is saved, it does not indicate whether or not caffeine is active
- The colours of the main menu item, as well as the systray icon, indicate whether caffeine is currently active or not

My problem with this design is that the menu items 5, 10, 30 infinite have two behaviours: they both set the time limit and they enable or disable caffeine mode. This makes the checkmark next to it confusing: does it refer to the time limit preference, or to the activation status of caffeine mode, or both? I was confused by this.

*Proposed design (design 2):*

This is the design that I suggested in my original comment, which I'm now abandoning

*Proposed design (design 3):*

[...]

PR #252, opened 2023-03-21
Timer behavior & Gnome 44 adjustment

pakaoraki:
### Timer behavior
To keep design consistent with official QuickSetting toggle like `Power Profiles`, I propose that Caffeine remember the last timer option selected (as we discussed in #234). 
It will be always a timer option selected: when changing duration range in settings, the selection will reset to `infinite`.

### Gnome 44: new Subtitle

Gnome 44 introduce a new subtitle label and last version of Caffeine already using it to show countdown when timer is enable.
With this change, it would also make sens to show which timer option is selected in toggle subtitle when Caffeine is disable. Subtitle is disable when `infinite` is selected.

Caffeine_timer_subtitle

I also think it can be great to see which app is enabling Caffeine:

Caffeine_app_subtitle

When having multiple active apps that trigger Caffeine, it will show only the last opened. Then, the name will be updated or removed as apps are closed one by one.

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -391,9 +391,6 @@ class Caffeine extends QuickSettings.SystemIndicator {
             this._resetAppSignalId();
             this._updateAppEventMode();
         });
-        this.connect('destroy', () => {
-            this.quickSettingsItems.forEach(item => item.destroy());
-        });
 
         // Change user state on icon scroll event
         this._indicator.reactive = true;
@@ -1030,6 +1027,9 @@ class Caffeine extends QuickSettings.SystemIndicator {
         // Remove all inhibitors
         this._appInhibitedData.forEach((data, appId) => this.removeInhibit(appId));
         this._appInhibitedData.clear();
+
+        // Remove ToggleMenu
+        this.quickSettingsItems.forEach(item => item.destroy());
 
         // Disconnect from signals
         if (this._settings.get_boolean(FULLSCREEN_KEY)) {
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -215,7 +215,7 @@ class CaffeineToggle extends QuickSettings.QuickMenuToggle {
             if(timer[0] === 0) {
                 label = _('Infinite');
             } else {
-                label = parseInt(timer[durationIndex]) + 'm';
+                label = parseInt(timer[durationIndex]) + _(' minutes');
             }
             if (!label)
                 continue;
@@ -790,7 +790,7 @@ class Caffeine extends QuickSettings.SystemIndicator {
         if (ShellVersion >= 44 && !this._settings.get_boolean(TOGGLE_STATE_KEY)) {
             const timerDuration = this._settings.get_int(TIMER_KEY);
             this._caffeineToggle.subtitle = timerDuration !== 0 
-                ? parseInt(timerDuration) + "m" 
+                ? parseInt(timerDuration) + _(' minutes')
                 : null;
         }
     }
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -340,15 +340,6 @@ class Caffeine extends QuickSettings.SystemIndicator {
         this._timeWorkspaceRemove = null;
         this._timeAppUnblock = null;
 
-        // Init settings keys and restore user state
-        this._settings.reset(TOGGLE_STATE_KEY);
-        if (this._settings.get_boolean(USER_ENABLED_KEY) && this._settings.get_boolean(RESTORE_KEY)) {
-            this.toggleState();
-        } else {
-            // reset user state
-            this._settings.reset(USER_ENABLED_KEY);
-        }
-
         // Show icon
         this._manageShowIndicator();
 
@@ -367,6 +358,15 @@ class Caffeine extends QuickSettings.SystemIndicator {
         this._caffeineToggle = new CaffeineToggle();
         this.quickSettingsItems.push(this._caffeineToggle);
         this._updateTimerSubtitle();
+
+        // Init settings keys and restore user state
+        this._settings.reset(TOGGLE_STATE_KEY);
+        if (this._settings.get_boolean(USER_ENABLED_KEY) && this._settings.get_boolean(RESTORE_KEY)) {
+            this.toggleState();
+        } else {
+            // reset user state
+            this._settings.reset(USER_ENABLED_KEY);
+        }
 
         // Bind signals
         this._inhibitorAddedId = this._sessionManager.connectSignal(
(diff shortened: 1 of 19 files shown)


stuarthayhurst:
I definitely like these changes, haven't tested yet, but I will tomorrow.

I do think that maybe '15m' should be '15 minutes' (for example, also applies to the other durations). If this is changed, the timer durations inside the dropdown would need updating too. What do you think?

caffeine@patapon.info/extension.js
```diff
@@ -545,36 +554,31 @@ class Caffeine extends QuickSettings.SystemIndicator {
...
-            let timerDelay = (this._settings.get_int(TIMER_KEY) * 60);
+        // Get duration
+        let timerDelay = (this._settings.get_int(TIMER_KEY) * 60);
 
-            // Execute Timer only if duration isn't set on infinite time
-            if(timerDelay !== 0) {
-                let secondLeft = timerDelay;
-                this._showIndicatorLabel();
+        // Execute Timer only if duration isn't set on infinite time
+        if(timerDelay !== 0) {
```
stuarthayhurst, line 566:
```suggestion
        if (timerDelay !== 0) {
```

caffeine@patapon.info/extension.js
```diff
@@ -228,11 +226,18 @@ class CaffeineToggle extends QuickSettings.QuickMenuToggle {
             this._itemsSection.addMenuItem(item);
         }
         this.menuEnabled = TIMERS.length > 2;
+     
+        // Select active duration
+        if(resetDefault && this._settings.get_int(TIMER_KEY) !== 0) {
```
stuarthayhurst, line 231:
```suggestion
        if (resetDefault && this._settings.get_int(TIMER_KEY) !== 0) {
```

stuarthayhurst (commented):
Very minor, but could these use consistent spaces? I know the whole codebase doesn't use the same consistent spaces everywhere, but it's easier to fix later if more inconsistencies haven't been added.

stuarthayhurst:
Tested on both GNOME 44 and GNOME 43, all working on my systems. Looks much better with the 'minutes' change too, nice :)

stuarthayhurst (approved):
I'm happy with this, although I can't really comment on the app names as a subtitle, as I don't really use the fullscreen app auto-enable feature.

caffeine@patapon.info/extension.js
```diff
@@ -948,6 +979,13 @@ class Caffeine extends QuickSettings.SystemIndicator {
             this._manageScreenBlankState(true); // Allow blank screen
             this._manageNightLight(false, true);
             this.addInhibit(appId); // Inhibit app
+
+            // Uninhibit previous focused apps
+            this._appInhibitedData.forEach((data, id) => {
+                if (id !== appId) {
+                    this.removeInhibit(id);
+                }
```
eonpatapon, line 987: This is a bugfix right ? We were keeping inhibitors for apps that were previously in focus ?

caffeine@patapon.info/extension.js
```diff
@@ -766,6 +773,28 @@ class Caffeine extends QuickSettings.SystemIndicator {
...
+            const listAppId = this._appInhibitedData.keys();
+            let appId = id !== null ? id : listAppId.next().value;
+            if (appId !== undefined) {
+                let appInfo = Gio.DesktopAppInfo.new(appId);        
+                this._caffeineToggle.subtitle = appInfo.get_display_name();
+            }
+        }
+    }
+
+    // Add the timer duration selected as subtitle (=< Gnome 44)
```
eonpatapon, line 788:
```suggestion
    // Add the timer duration selected as subtitle (>= Gnome 44)
```

caffeine@patapon.info/extension.js
```diff
@@ -766,6 +773,28 @@ class Caffeine extends QuickSettings.SystemIndicator {
         }
     }
 
+    // Add the name of App as subtitle (=< Gnome 44)
```
eonpatapon, line 776:
```suggestion
    // Add the name of App as subtitle (>= Gnome 44)
```

caffeine@patapon.info/extension.js
```diff
@@ -228,11 +226,18 @@ class CaffeineToggle extends QuickSettings.QuickMenuToggle {
             this._itemsSection.addMenuItem(item);
         }
         this.menuEnabled = TIMERS.length > 2;
+     
+        // Select active duration
```
eonpatapon, line 230: What's the logic here ? Reset but not always ? :)

caffeine@patapon.info/extension.js
```diff
@@ -545,36 +554,31 @@ class Caffeine extends QuickSettings.SystemIndicator {
...
-            this._timerEnable = true;
-
-            // Reset Timer
-            this._removeTimer(true);
+        this._timerEnable = true;
 
-            // Enable Caffeine
-            this._settings.set_boolean(TOGGLE_STATE_KEY, true);
+        // Reset timer
+        this._removeTimer(true);
```
eonpatapon, line 560: I believe if you call first `_removeTimer()` before `this._timerEnable = true;` you don't need to handle a `reset` param in `_removeTimer()`

caffeine@patapon.info/extension.js
pakaoraki: I need to call `_sync()` at this step, this function is also called when `TIMER_KEY` changed. I wanted to avoid calling two time `_sync()` in case we reset timer to 0.

caffeine@patapon.info/extension.js
pakaoraki: Yes, you're right. I'll rewrite this. :-)

caffeine@patapon.info/extension.js
pakaoraki: Yes, before we were keeping previously inhibitors for apps in focus mode, but as soon as user focus a none trigger app, then all of them were removed.
With the new subtitle label that show which app is enabling Caffeine, I wanted to update the app name when focusing two trigger app one by one. For that, I need to remove the previous inhibitor.

I just realize that the 'user' inhibitor is also removed and that's not good because when having an active timer and focusing an app, it disable caffeine...
I will write a separate commit to fix this situation.

caffeine@patapon.info/extension.js
pakaoraki: It reset only when duration range changes in settings, but not when the extension starts. So this way, it keep the last option selected when Caffeine restart.

pakaoraki:
After more testing I found another bug: when disable the extension while timer is enable, it crash. Will fix it asap.

pakaoraki:
> After more testing I found another bug: when disable the extension while timer is enable, it crash. Will fix it asap.

It's fixed

stuarthayhurst:
Is there anything left to sort, or is this ready for review?

pakaoraki:
It's ready for review. I just add this last commit to fix this issue: #254.
I tested it on Fedora 38/Ubuntu 23.04 and it run fine.

eonpatapon (approved):

PR #258, opened 2023-04-15
Update CLI documentation

### LINKED ISSUE #257: CLI options do not work
When I want to enable Caffeine from command line, it does not work, when I get the value, it shows proper true/false, however, when I change it via command, I get the desired true/false via command, but it does not control the actual Caffeine on/off state (icon is still the same, no changes).

pakaoraki:
Will fix #257.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,15 +64,15 @@ This option enable Caffeine for a given amount of time (similar to Caffeine in L
 
 - Get current state:
   ```sh
-  gsettings --schemadir ~/.local/share/gnome-shell/extensions/caffeine@patapon.info/schemas/ get org.gnome.shell.extensions.caffeine user-enabled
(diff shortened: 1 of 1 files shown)


stuarthayhurst (approved):

stuarthayhurst:
Is `user-enabled` just used internally then?

pakaoraki:
Yes, it's used to store the user state of Caffeine (state toggled by user, not by apps) for restoring the previous Caffeine state (with "Remember state" option).

```
this._settings.reset(TOGGLE_STATE_KEY);
if (this._settings.get_boolean(USER_ENABLED_KEY) && this._settings.get_boolean(RESTORE_KEY)) {
         this.toggleState();
} else {
         // reset user state
         this._settings.reset(USER_ENABLED_KEY);
}
```

PR #260, opened 2023-04-18
Add quick-and-dirty support for toggling xscreensaver

greatquux:
add quick-and-dirty support to toggle xscreensaver along with caffeine

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -45,12 +45,15 @@ const TRIGGER_APPS_MODE = 'trigger-apps-mode';
 const INDICATOR_POSITION = 'indicator-position';
 const INDICATOR_INDEX = 'indicator-position-index';
 const INDICATOR_POS_MAX = 'indicator-position-max';
+const XSCREENSAVER_STOP = "xscreensaver-command -exit";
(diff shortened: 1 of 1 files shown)


greatquux:
xscreensaver can be configured to put monitors to sleep (or not), just works like the other ones, except some of us like the cute animations, etc. Yeah it should definitely be put behind a user setting defaulting to off, I am just not knowledgeable enough to do that part of it.

eonpatapon:
This extension doesn't control directly screensavers or specific apps. It designed to manage inhibitors through gnome session manager which in turn will create systemd inhibitors.

Once the inhibitor is in place, applications should take it in account.

I'm reluctant to add support for specific applications that ignore inhibitors. I guess xscreensaver doesn't support systemd inhibitors but I found some related daemon that probably does what you want: https://man.archlinux.org/man/xscreensaver-systemd.6.en

So I'd suggest to just use this, and this won't require any change in this extension.

pakaoraki:
> Once the inhibitor is in place, applications should take it in account.

I agree.

greatquux:
Ah, very interesting - I will look into that! I think it was being started but wasn't working for whatever reason. In any case yeah I'll close this because at only 4 lines it's kind of simple to rebase it anyway.

PR #246, opened 2023-03-13
Update gnome-shell-extension-caffeine.po for zh_CN

xiaozhangup:

PR #262, opened 2023-04-19
Translate 'Caffeine' in QuickSettings

pakaoraki:
I noticed that in Lineageos they translated the menu 'Caffeine', so I thought it could make sens to make this change. 

Also, someone in #246 had this idea:

> I wonder if i should translate word "Caffeine"?

Let me know if it's something you'd like to see.

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -98,8 +98,6 @@ const DBusSessionManagerInhibitorIface = '<node>\
 const DBusSessionManagerInhibitorProxy = Gio.DBusProxy.makeProxyWrapper(DBusSessionManagerInhibitorIface);
 
 const IndicatorName = 'Caffeine';
-const IndicatorLabel = _('Caffeine');
-const TimerMenuName = _('Caffeine timer');
 const DisabledIcon = 'my-caffeine-off-symbolic';
 const EnabledIcon = 'my-caffeine-on-symbolic';
 const TimerMenuIcon = 'stopwatch-symbolic';
@@ -159,7 +157,7 @@ class CaffeineToggle extends QuickSettings.QuickMenuToggle {
             // The 'label' property was renamed to 'title' in GNOME 44 but quick settings have otherwise 
             // not been changed. The below line allows support for both GNOME 43 and 44+ by using the 
             // appropriate property name based on the GNOME version.
-            [ShellVersion >= 44 ? 'title' : 'label']: IndicatorLabel,
+            [ShellVersion >= 44 ? 'title' : 'label']: _('Caffeine'),
             toggleMode: true,
         });
 
@@ -176,7 +174,7 @@ class CaffeineToggle extends QuickSettings.QuickMenuToggle {
         this._iconName();
 
         // Menu
-        this.menu.setHeader(this.finalTimerMenuIcon, TimerMenuName, null);
+        this.menu.setHeader(this.finalTimerMenuIcon, _('Caffeine timer'), null);
 
         // Add elements
         this._itemsSection = new PopupMenu.PopupMenuSection();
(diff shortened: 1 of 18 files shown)


stuarthayhurst:
On one hand, it's the name of the extension, but it's also a reference to actual caffeine of course, which would make less sense in other languages, unless it's translated.

I guess I'm for this, but I have no strong feelings either way.

caffeine@patapon.info/extension.js
```diff
@@ -98,7 +98,6 @@ const DBusSessionManagerInhibitorIface = '<node>\
 const DBusSessionManagerInhibitorProxy = Gio.DBusProxy.makeProxyWrapper(DBusSessionManagerInhibitorIface);
 
 const IndicatorName = 'Caffeine';
-const TimerMenuName = _('Caffeine timer');
```
stuarthayhurst, line 101: I'm not sure I understand the motivation for this? This removes the one instance of the translation, instead translating each time the label is created / updated.

caffeine@patapon.info/extension.js
```diff
@@ -158,7 +157,7 @@ class CaffeineToggle extends QuickSettings.QuickMenuToggle {
             // The 'label' property was renamed to 'title' in GNOME 44 but quick settings have otherwise 
             // not been changed. The below line allows support for both GNOME 43 and 44+ by using the 
             // appropriate property name based on the GNOME version.
-            [ShellVersion >= 44 ? 'title' : 'label']: IndicatorName,
```
stuarthayhurst, line 161: This still leaves an instance of IndicatorName, which is then untranslated. Was this intentional? If the goal was to do the same thing as `TimerMenuName`, then I don't understand the reasoning, for the same reason as my last comment.

stuarthayhurst (commented):
This also needs to be rebased, as the translations have changed, as well as `extensions.js` ('Caffeine timer' -> 'Caffeine Timer')

caffeine@patapon.info/extension.js
stuarthayhurst: Unless the intention was to not translate the remaining instance of `IndicatorName`, but if that's the case, I think the constant should probably be removed, as there's only 1 instance of it left, and it's an unchanging string.

caffeine@patapon.info/extension.js
pakaoraki: For some reasons, the translation did not work on my previous test when I was using constants `IndicatorLabel` and `TimerMenuName`, so I simply removed them. I will update the code and redo some test.

caffeine@patapon.info/extension.js
pakaoraki: `IndicatorName` is still used as a generic name for the inhibit flag only (so it will stay untranslated):
```
addInhibit(appId) {
        this._sessionManager.InhibitRemote(appId,
            0, 'Inhibit by %s'.format(IndicatorName), this.inhibitFlags,
            cookie => {
```
We can removed this constant and maybe we could also get this name from the json file with `Me.metadata.name` instead.

stuarthayhurst: If it doesn't need to be translated, why can't we just hard-code it to say 'Inhibit by Caffeine'?

pakaoraki: I personally prefer to avoid hard coding, but I'm maybe be old school. The name of the extension is define in the metadata.json so I reckon it would be cleanest to use this instead.

pakaoraki:
Since the master branch changed, I will close this PR and come back with new one:

- I will propose in a first time an updated french translation.
- Maybe come back in a second time with a dedicated PR for translating `Caffeine` and cleaning some code.

caffeine@patapon.info/extension.js
stuarthayhurst: That's understandable, but since the code is for this extension and not a generic piece of code dropped into multiple extensions, it makes sense to hard-code the string, instead of just querying the extensions system to retrieve another hard-coded string.

Ultimately, if it's cleanly done I'm not really opposed to it, I started this comment thread to try get at the reasoning for some of the changes, that's all :)

PR #273, opened 2023-06-28
Update Hungarian translation

infeeeee:


stuarthayhurst:
Sorry, I messed up and the translation files are slightly out of date
I'll push the corrected ones and this will probably need a rebase

infeeeee:
Yeah, something was fishy for me. But I compared to other languages, and the strings were the same. Let me know when  I should check again.

stuarthayhurst:
`master` has the correct files now, sorry about that. You should be good to check again now, there shouldn't be a huge difference between the old and new versions.

stuarthayhurst:
`master` has changed significantly, it's now a different file you have to update, you should find it as `caffeine@patapon.info/locale/hu.po`.

pakaoraki:
Replaced by #291

PR #277, opened 2023-07-09
Translate 'Caffeine' in all languages

pakaoraki:
I noticed that, depend of the language, sometimes 'caffeine' is translated, sometimes not: it could be nice to translate it in all to keep consistency.
Also in Lineageos they translated the menu 'Caffeine', so I thought it could make sens to make this change.

Someone in [#246](https://github.com/eonpatapon/gnome-shell-extension-caffeine/pull/246) had this idea:

> I wonder if i should translate word "Caffeine"?

Let me know if it's something you'd like to see.

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -157,7 +157,7 @@ class CaffeineToggle extends QuickSettings.QuickMenuToggle {
             // The 'label' property was renamed to 'title' in GNOME 44 but quick settings have otherwise
             // not been changed. The below line allows support for both GNOME 43 and 44+ by using the
             // appropriate property name based on the GNOME version.
-            [ShellVersion >= 44 ? 'title' : 'label']: IndicatorName,
+            [ShellVersion >= 44 ? 'title' : 'label']: _('Caffeine'),
             toggleMode: true
         });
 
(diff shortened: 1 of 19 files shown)


pakaoraki:
I just realized that the Korean file is `ka.po`, I think it suppose to be `ko.po`.

`caffeine@patapon.info/locale/ko/LC_MESSAGES/gnome-shell-extension-caffeine.po → caffeine@patapon.info/locale/ka.po`
I can rename this file in this PR if this change is accepted.

stuarthayhurst:
> I just realized that the Korean file is `ka.po`, I think it suppose to be `ko.po`.
> 
> `caffeine@patapon.info/locale/ko/LC_MESSAGES/gnome-shell-extension-caffeine.po → caffeine@patapon.info/locale/ka.po` I can rename this file in this PR if this change is accepted.

I mean ideally we'd have separate PRs for separate changes to keep rebases simple and following changes, but I suppose it doesn't matter too much for a small change like this

But yes, `ka` seems to be the code for Georgian, that must've been a typo when I moved the translation files, thanks for catching that

stuarthayhurst:
Translating Caffeine in all languages would make sense, as English 'Caffeine' doesn't really have any meaning in other languages.

stuarthayhurst:
Is it worth combining the last 2 commits? Ideally we wouldn't have 2 commits to update the locales.

pakaoraki:
I combined the last 2 commits

stuarthayhurst (approved):
Thanks :)

PR #280, opened 2023-07-18
Fix Gdk.SHIFT_MASK being undefined

### LINKED ISSUE #279: FYI: Are you sure to use "Gdk.SHIFT_MASK" in generalPage.js
I have similar code and I found "`Gdk.SHIFT_MASK`" was undefined while debugging it under my environment of GNOME43.

In this code `generalPage.js` file uses this.  I think desired constant is "`Gdk.ModifierType.SHIFT_MASK`", instead.

You can test situation for your build environment using my test code under `test-keys/*` of https://github.com/osamuaoki/inputmethod-shortcuts

(I don't know if this issue is version dependent.  I see similar code as yours every where.  E.g., https://github.com/jqno/gnome-happy-appy-hotkey/)

stuarthayhurst:
Bindings such as "SHIFT" + "A" were mistakenly being enabled. I'm not sure if / when it swapped to `Gdk.ModifierType.SHIFT_MASK`, instead of `Gdk.SHIFT_MASK`, but it's the same on both GNOME 43 and 44, the 2 supported versions.

Closes #279

### DIFF
--- a/caffeine@patapon.info/preferences/generalPage.js
+++ b/caffeine@patapon.info/preferences/generalPage.js
@@ -371,7 +371,7 @@ const ShortcutSettingWidget = class extends Adw.ActionRow {
     }
 
     isValidBinding(mask, keycode, keyval) {
-        return !(mask === 0 || mask === Gdk.SHIFT_MASK && keycode !== 0 &&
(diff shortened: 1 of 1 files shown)


pakaoraki (approved):

pakaoraki:
Tested on Fedora 38, working as expected. Thanks

PR #281, opened 2023-07-24
Update metadata.json to gnome 45

gbPagano:
just adding gnome 45 in metadata.json, tested and working.

### DIFF
--- a/caffeine@patapon.info/metadata.json
+++ b/caffeine@patapon.info/metadata.json
@@ -2,7 +2,8 @@
   "version": 48,
   "shell-version": [
     "43",
-    "44"
+    "44",
+    "45"
   ],
   "uuid": "caffeine@patapon.info",
   "name": "Caffeine",


stuarthayhurst:
Sorry, I can't approve this, the extension will need large changes for GNOME 45.

Upstream, many large changes are planned / implemented:
 - ESM modules, breaking every extension due to syntax changes
 - Preferences handling (new classes / init methods?)
 - Translation / boilerplate initialisations (inherit from new classes, and the shell will handle it)
 - Extension class changes (standard extension class to inherit, I believe this will be mandatory to pass review)
 - Split up ExtensionUtils (it's becoming 4 separate modules now)

libadwaita 1.4, an injection manager and other miscellaneous changes are all coming too, but don't directly impact this extension.

Many of these changes are still in pull request form or in `git` and not in a tagged build, but are all planned for GNOME 45. The upstream devs have planned GNOME 45 as a release where they more or less break every extension, so they can get all breaking changes through at once, instead of breaking extensions several times over multiple releases.

I haven't started porting work yet, as the changes are still very much WIP, but when the changes are merged and I have time, I'll look at porting the extension properly.

Anyway, thanks for testing :)

PR #286, opened 2023-08-11
Fix multiple Fullscreen inhibit

### LINKED ISSUE #285: Fullscreen inhibit gets stuck due to a race condition
Steps to reproduce:

1. Configure Caffeine to inhibit screen lock for fullscreen windows (the default configuration).
2. Open and close a fullscreen window very frequently. For example, open a picture in Eog and double-click on it constantly to toggle fullscreen.
3. Observe a lot of "fullscreen" inhibitors accumulate using the following script:
```
busctl --user -j call org.gnome.SessionManager /org/gnome/SessionManager org.gnome.SessionManager GetInhibitors | jq -r '.data[0][]' | while read -r obj; do
	busctl --user call org.gnome.SessionManager "$obj" org.gnome.SessionManager.Inhibitor GetAppId
done
```
Output:
```
s "fullscreen"
s "fullscreen"
s "fullscreen"
s "fullscreen"
s "fullscreen"
s "fullscreen"
s "fullscreen"
s "fullscreen"
s "fullscreen"
s "fullscreen"
```

By adding debug prints, I found out that `addInhibit('fullscreen')` can be called multiple times in a row, followed by only one `removeInhibit('fullscreen')`. What I believe to be the root cause of the issue is that `addInhibit()` doesn't populate `this._appInhibitedData` right away, but only after the callback is called. Combined with a 2-second delay in `toggleFullscreen()`, this means that this [piece of code](https://github.com/eonpatapon/gnome-shell-extension-caffeine/blob/31b00d945add65ad5b19900d451d13f25978be85/caffeine%40patapon.info/extension.js#L438-L441) may be scheduled multiple times over a 2-second window, and when it comes to executing it, `this._appInhibitedData.has('fullscreen')` will return false every time, `addInhibit` will be called more than once, and only then its callback will kick in and populate `this._appInhibitedData`, effectively overwriting old cookies and not allowing to remove the inhibit anymore.

pakaoraki:
Fix #285

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -433,13 +433,18 @@ class Caffeine extends QuickSettings.SystemIndicator {
     }
 
     toggleFullscreen() {
-        // Reset previous FullScreen delay
+        /* Reset previous FullScreen delay
+        * This prevent multiple inhibitors to be created in toggleFullscreen()
+        * if a previous timer is still running.
+        */
         if (this._timeFullscreen !== null) {
             GLib.Source.remove(this._timeFullscreen);
             this._timeFullscreen = null;
         }
 
         this._manageScreenBlankState(false);
+
+        // Add 2 second delay before adding inhibitor
         this._timeFullscreen = GLib.timeout_add_seconds(GLib.PRIORITY_DEFAULT, 2, () => {
             if (this.inFullscreen && !this._appInhibitedData.has('fullscreen')) {
                 this.addInhibit('fullscreen');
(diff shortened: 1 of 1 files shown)


stuarthayhurst:
Maybe we should add something about the race condition to the comment? Then it's super obvious why we do that.

Tested on my system, works perfectly :)

stuarthayhurst (approved):

pakaoraki:
I updated the comment: let me know i you want to modify or change something.

PR #287, opened 2023-08-11
Turkish translation updated

osmancoskun:
- Half of translations were missing.
- Turkish translation updated for future usage.


stuarthayhurst:
Thanks :)

PR #288, opened 2023-08-14
FIX: missing translations

osmancoskun:
There was a small bug. In my current computer, some translations wasnt working even if i had translated these words on `.po` file. after making some research on codes, I found that `gettext` doesnt work if its assigned to a variable. I just changed these variables to normal `gettext` function `_('some words')`
here is how it is before fix:

image

here is how it seems after fix

Ekran Görüntüsü - 2023-08-14 13-53-52


stuarthayhurst (changes requested):
Did you change the indent for any specific reason? It was done that way for consistency with the rest of the extension, and also to match the upstream style in GNOME.

It's also made it much harder to see what this PR actually changes, as it's buried in a sea of indent changes.

Still, good spot with the translations, thanks :)

osmancoskun:
sorry for that indent. it was done by my extension probably. i will change it asap

Issue #218, opened 2022-12-17
RHEL 9.1 error


srosciszewski:
Technical details:

SyntaxError: expected property name, got '{'

Stack trace:
  _init@resource:///org/gnome/Shell/Extensions/js/extensionsService.js:216:33
  OpenExtensionPrefsAsync/<@resource:///org/gnome/Shell/Extensions/js/extensionsService.js:128:33
  asyncCallback@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:115:22
  run@resource:///org/gnome/Shell/Extensions/js/dbusService.js:177:20
  main@resource:///org/gnome/Shell/Extensions/js/main.js:19:13
  run@resource:///org/gnome/gjs/modules/script/package.js:206:19
  start@resource:///org/gnome/gjs/modules/script/package.js:190:8
  @/usr/share/gnome-shell/org.gnome.Shell.Extensions:1:17

pakaoraki:
I can confirm the issue: the extension work as expected, but we get this error when opening preferences. RHEL 9.1 use gnome 40.4 and Caffeine v42.

pakaoraki:
The problem come from this [commit](https://github.com/eonpatapon/gnome-shell-extension-caffeine/commit/26cbfb89aa910e32ec25e356c568259016840b7c#diff-c34f5a08bb4dc9f07bc07f46a87d92b7e7c17dd5521cd1b8709080d9950230a8). it seems like the `ShortcutSettingWidget` cause this error. I can't really find the root of the problem for the moment. Also, I reckon that using `Adw` in gnome 40 would generate another crash: `Error: Requiring Adw, version none: Typelib file for namespace 'Adw' (any version) not found.`

So maybe the best thing to do is simply removing this feature for gnome 40. So, create a specific fixed branch for gnome 40 and keep the current v42 for gnome 41/42. What do you think @eonpatapon ?

pakaoraki:
Hi @srosciszewski 

I fix this issue and I test it successfully in RHEL 9.1 / gnome 40. Could you please confirm that it work for you ?

Remove your current version of Caffeine, then:
```
git clone --branch gnome-shell-40 https://github.com/pakaoraki/gnome-shell-extension-caffeine.git
cd gnome-shell-extension-caffeine
make build
make install
```

After logout, preferences should be accessible again.

stuarthayhurst:
[...]

PR #293, opened 2023-08-27
Fix for RHEL9 - Gnome40

pakaoraki:
This will fixe #218 .
@eonpatapon @stuarthayhurst, I'm not sure about publishing a new release for Gnome 40, but this dedicated branch can be archived for people having this issue.


stuarthayhurst:
Will the version to branch table in the README of the master branch need updating?

pakaoraki:
> Will the version to branch table in the README of the master branch need updating?

We can add a note in the README if this PR is accepted.

> I'll leave release management to eon, but I personally wouldn't worry about a release for it. If anyone is still running that, they're welcome to come and manually install, but I don't think it's worth making releases for old versions.

I do agree with you, we should not maintain Caffeine for older version of gnome and we don't need to publish a new release, but I feel bad for the few users that run Caffeine in RHEL9: it can be nice to leave this dedicated branch for manual install.

eonpatapon:
I think I can upload a new version with this, it will target gnome 40 only

eonpatapon:
Uploaded on e.g.o

PR #294, opened 2023-09-03
po: correct Catalan translation

juxuanu:
https://github.com/eonpatapon/gnome-shell-extension-caffeine/pull/277/commits/af19c1148dd82888417be2803736bcdeb7995150 introduced a regression on the translation.


stuarthayhurst:
Thanks

PR #298, opened 2023-09-26
Complete spanish translations

delthia:
Hi,
I've completed the Spanish translations for this project, however, here are some things I'd like to mention:

- I've changed the name of the project in the Spanish translations to not be translated. Before in the translations *caffeine* had been translated to Spanish but, as it is the project name's, I think it is more appropriate to have it maintain its original name.
-  There is a translation that is missing, and some that I think that could be more clearly translated. If I can think of them, I'll add the change. In case the PR is approved and I later come up with better translations, I'll be sure to contribute again

Thanks

### DIFF
--- a/caffeine@patapon.info/locale/es.po
+++ b/caffeine@patapon.info/locale/es.po
@@ -106,37 +106,30 @@ msgid "Never"
 msgstr "Nunca"
 
 #: preferences/displayPage.js:52
-#, fuzzy
 msgid "Show status indicator in top panel"
 msgstr "Mostrar el indicador de Cafeína en el panel superior"
 
 #: preferences/displayPage.js:53
-#, fuzzy
 msgid "Enable or disable the Caffeine icon in the top panel"
 msgstr "Activa o desactiva el icono de Cafeína en el panel superior"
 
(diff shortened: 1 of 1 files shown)


stuarthayhurst:
#277 changed the translation behaviour of the project name, I believe the move was for consistency, and that 'Caffeine' doesn't really have meaning in other languages, the same way that in English it relates to keeping awake.

Anyway, thanks for the updates, if you think there's more to do I won't merge this yet

delthia:
Oh, well, I didn't see that issue. It makes sense, so I'll translate that in a bit and add the commit.

delthia:
I think it should be correctly translated now

stuarthayhurst:
Thanks. Can you take a look at the translations marked as fuzzy please? If they're all good they just need the fuzzy tag removed.

delthia:
I've now removed the fuzzy tag from the translations I'm sure they are correct.
There are some that still have it as while I think they are "understandable", are not necessarily the best they could be.

delthia:
I'd say that this is complete and ready for merging.

Thanks

PR #301, opened 2023-10-09
Add missing metadata extension name

pakaoraki:
This is a small fix for the missing extension name when creating **Caffeine**. It was resulting in bad inhibitor description: `Inhibit by undefined` instead of `Inhibit by Caffeine`. It take the name extension from the metadata file.

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -1106,13 +1106,6 @@ class Caffeine extends QuickSettings.SystemIndicator {
 });
 
 export default class CaffeineExtension extends Extension {
-    /**
-     * @param {ExtensionMeta} metadata - Caffeine extension metadata
-     */
-    constructor(metadata) {
-        super(metadata);
-    }
-
     enable() {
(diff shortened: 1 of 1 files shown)


stuarthayhurst:
Not sure how I missed this, thanks

caffeine@patapon.info/extension.js
```diff
@@ -1106,9 +1106,16 @@ class Caffeine extends QuickSettings.SystemIndicator {
 });
 
 export default class CaffeineExtension extends Extension {
+    /**
+     * @param {ExtensionMeta} metadata - Caffeine extension metadata
+     */
+    constructor(metadata) {
+        super(metadata);
+    }
+
```
stuarthayhurst, line 1115:
```suggestion

```

Do we really need this? We aren't overriding `constructor()`, I don't think we need to do anything here.

pakaoraki: Absolutely not :-) sorry about that...

stuarthayhurst (approved):

PR #303, opened 2023-10-15
Sync number version with release - v52 - Gnome 45

pakaoraki:
There is a gap between the version number present in Github and the release one : this can be a bit confusing. 
This fix will update the version to the current release v52 (master branch).


eonpatapon:
Yeah probably due to a re-upload

PR #310, opened 2023-12-12
Update German translation

daPhipz:
I noticed there are also branches for [`gnome-shell-40`](https://github.com/eonpatapon/gnome-shell-extension-caffeine/tree/gnome-shell-40) and [`gnome-shell-43-44`](https://github.com/eonpatapon/gnome-shell-extension-caffeine/tree/gnome-shell-43-44) - should I file a PR against these branches as well, or is the development for these branches finished?

### DIFF
--- a/caffeine@patapon.info/locale/de.po
+++ b/caffeine@patapon.info/locale/de.po
@@ -2,50 +2,50 @@ msgid ""
 msgstr ""
 "Project-Id-Version: gnome-shell-extension-caffeine\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-08-15 00:10+0200\n"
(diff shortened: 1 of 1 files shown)


stuarthayhurst:
> I noticed there are also branches for [`gnome-shell-40`](https://github.com/eonpatapon/gnome-shell-extension-caffeine/tree/gnome-shell-40) and [`gnome-shell-43-44`](https://github.com/eonpatapon/gnome-shell-extension-caffeine/tree/gnome-shell-43-44) - should I file a PR against these branches as well, or is the development for these branches finished?

I don't see why not since users may install from those branches, but I don't think it'll make it to an uploaded release, it'll just stay in the branch.

Either way, thanks for this PR :)

PR #315, opened 2024-02-13
Drop old and unused code

stuarthayhurst:
Found a line from around GNOME 3.28 in there, and looking at `this._screen` and `this._display`, they were both just `global.display`, also also tied to an unused signal.


eonpatapon (approved):

pakaoraki:
> we can drop the variables to track IDs, and drop the null checks when we disconnect, since that's all handled

That's a nice simplification :)

PR #316, opened 2024-02-14
Move from connect() to connectObject() where applicable

stuarthayhurst:
`connectObject()` allows tying the lifetime of the connection to the lifetime of another object. If the connected object or the object its lifetime is tied to is destroyed, the connection is destroyed (uses the `destroy` signal).

This means in several cases we can just use `connectObject()` to tie the lifetime to `this`, and we don't need to track any signal IDs, as its destruction will be taken care of for us. Another benefit is that we can connect multiple signals in 1 call, by repeating `'signal', handler` each time as arguments, instead of duplicating the call.

Any unused signals were also removed :)


stuarthayhurst:
#317 includes this, as it was branched from this branch
Closing in favour of that instead

PR #317, opened 2024-02-18
Port extension to GNOME 46

stuarthayhurst:
Self-explanatory, ports the extension to GNOME 46. The indicators don't have actors anymore, so `remove_child` is used instead.

Marked as a draft since the extension needs to be disabled and re-enabled to get the indicator position setting to work.

Since it makes sense to make a release after this gets merged, I updated the README version table for the next version, with GNOME 46 support

Branched from #316

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -29,10 +29,12 @@ import GLib from 'gi://GLib';
 
 import { Extension, gettext as _ } from 'resource:///org/gnome/shell/extensions/extension.js';
 
+import * as Config from 'resource:///org/gnome/shell/misc/config.js';
 import * as Main from 'resource:///org/gnome/shell/ui/main.js';
 import * as PopupMenu from 'resource:///org/gnome/shell/ui/popupMenu.js';
 import * as QuickSettings from 'resource:///org/gnome/shell/ui/quickSettings.js';
 const QuickSettingsMenu = Main.panel.statusArea.quickSettings;
+const ShellVersion = parseFloat(Config.PACKAGE_VERSION);
 
 const INHIBIT_APPS_KEY = 'inhibit-apps';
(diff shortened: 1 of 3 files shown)


stuarthayhurst:
I think the extension needs to be disabled and re-enabled because when it starts up the other indicators don't exist yet, so it thinks there aren't any, setting the max pos to -1, effectively disabling the setting.

stuarthayhurst:
Everything seems to be working on GNOME 45 and GNOME 46 now

caffeine@patapon.info/extension.js
```diff
@@ -903,13 +931,13 @@ class Caffeine extends QuickSettings.SystemIndicator {
 
     _appWorkspaceChanged() {
         // Reset signal for Add/remove windows on workspace
-        if (this._appAddWindowSignalId > 0) {
```
pakaoraki, line 906: In the first place, I bug when I saw my old code because I didn't understand why I chose to compare to 0. Then I remember that using "null" value was causing log error "invalid (NULL) pointer instance" or something like that. 

So I just ran few tests with your changes and this time there is no more errors in the logs when switching workspace:  maybe because of a change/fix upstream since I wrote this code, or maybe I just miss something :-). Anyway, it's working fine this time and it's better like this, thanks.

stuarthayhurst: Strange, I've never run into anything like that before, I just changed this to be a bit more literal, since we're really checking that it's for a signal, not specifically that it's above 0.

stuarthayhurst:
@eonpatapon GNOME 46 is due in just over a week, is there any chance this can be merged and released as `v53` soon?

This release should also sort the issue that the latest release is tagged as `v50` instead of `v52`, which is causing the wrong version to be reported as the latest in Debian's package tracker

Thanks :)

eonpatapon (approved):

PR #319, opened 2024-03-14
Add cutomizable timer duration

### LINKED ISSUE #296: longer timers
The default timers seems to have very low values. The shortest one is 5 minutes. Now, if I wanted to keep computer awake for 5 more minutes, wasn't it enough to move the mouse?

My usecase for caffeine is listening to music and leaving the computer unattended. In that scenario, what I need is something like 2h caffeine mode, while the longest one is 30 minutes.

Since there is no way to customize the duration, can you at least increase the default timers?

pakaoraki:
Add new "Timer" menu in preferences with customizable timer duration.

Screenshot from 2024-03-14 18-07-07 resized

This will close #296


pakaoraki:
It look like there is some issues with my code, I can't fix PR right now, I will have more time tomorrow evening. Cheers

stuarthayhurst:
I haven't looked at the code yet, but would it be better to store timer lengths using 1 key as an array, instead of 3 separate keys?

Also, if I remember correctly, red is supposed to be for destructive actions according to the GNOME Human Interface Guidelines I think it was? I don't know if that applies to extensions, but it might be worth considering to be consistent with the rest of GNOME.

pakaoraki:
> I haven't looked at the code yet, but would it be better to store timer lengths using 1 key as an array, instead of 3 separate keys?

Yes I totally agree, I was thinking the same. I just used a simple 'int' type for each duration but it need to be unified with a single key. 

> Also, if I remember correctly, red is supposed to be for destructive actions according to the GNOME Human Interface Guidelines I think it was? I don't know if that applies to extensions, but it might be worth considering to be consistent with the rest of GNOME.

It make sense, I hesitated to keep the red color anyway so I can remove it.

There are some other small improvements to make, I'll try to fix it all this evening or tomorrow for final review.

pakaoraki:
There is a weird eslint issue:

> /home/runner/work/gnome-shell-extension-caffeine/gnome-shell-extension-caffeine/caffeine@patapon.info/preferences/timerPage.js
>   283:13  error  Expected property shorthand  object-shorthand

The code is the same as line 66, but it cause an error only in line 283. Should I ignore this using `/* eslint-disable object-shorthand */` and `/* eslint-enable object-shorthand */` or did I miss something ?

stuarthayhurst:
> There is a weird eslint issue:
> 
> > /home/runner/work/gnome-shell-extension-caffeine/gnome-shell-extension-caffeine/[caffeine@patapon.info](mailto:caffeine@patapon.info)/preferences/timerPage.js
> > 283:13  error  Expected property shorthand  object-shorthand
> 
> The code is the same as line 66, but it cause an error only in line 283. Should I ignore this using `/* eslint-disable object-shorthand */` and `/* eslint-enable object-shorthand */` or did I miss something ?

I think it's because the variables have different names, `durationIndex` and `value`. I think changing the name of `value`, or replacing `value: value` with `value` should do it?

pakaoraki:
Thanks

caffeine@patapon.info/extension.js
```diff
@@ -100,6 +100,12 @@ const DBusSessionManagerInhibitorProxy = Gio.DBusProxy.makeProxyWrapper(DBusSess
 const DisabledIcon = 'my-caffeine-off-symbolic';
 const EnabledIcon = 'my-caffeine-on-symbolic';
 const TimerMenuIcon = 'stopwatch-symbolic';
+const timerIcons = [
```
stuarthayhurst, line 103:
```suggestion
const TimerIcons = [
```

If this is a global constant, should it use a capital instead?

caffeine@patapon.info/extension.js
```diff
@@ -182,27 +181,38 @@ class CaffeineToggle extends QuickSettings.QuickMenuToggle {
...
-
-        for (const timer of TIMERS) {
+        const variantDuration = this._settings.get_value(DURATION_TIMER_LIST);
+        const durationValues = variantDuration.deepUnpack();
+        const shortTimer = durationValues[0];
+        const mediumTimer = durationValues[1];
+        const longTimer = durationValues[2];
+
+        // Create menu timer
+        for (const [index, timer] of [shortTimer, mediumTimer, longTimer, 0].entries()) {
```
stuarthayhurst, line 191: Can `shortTimer`, `mediumTimer` and `longTimer` be dropped, and instead use `[...durationValues, 0].entries()`?

I don't use the spread operator much, so I may be wrong, but this seems like a good opportunity for it

pakaoraki: Yes, I can remove these constants and simplify this part.

caffeine@patapon.info/preferences/timerPage.js
```diff
@@ -0,0 +1,346 @@
...
+        let enableCustomTimerSwitch = new Gtk.Switch({
+            valign: Gtk.Align.CENTER,
+            active: this._settings.get_boolean(this._settingsKey.USE_CUSTOM_DURATION)
+        });
+        this.enableCustomTimerRow = new Adw.ActionRow({
+            title: _('Enable custom values'),
+            subtitle: _('Select custom value for each duration'),
+            activatable_widget: enableCustomTimerSwitch
+        });
+        this.enableCustomTimerRow.add_suffix(enableCustomTimerSwitch);
```
gabrbrand, line 134:
```suggestion
        this.enableCustomTimerRow = new Adw.SwitchRow({
            title: _('Enable custom values'),
            subtitle: _('Select custom value for each duration'),
            active: this._settings.get_boolean(this._settingsKey.USE_CUSTOM_DURATION)
        });
```
[SwitchRow](https://gjs-docs.gnome.org/adw1~1/adw.switchrow) was introduced with libadwaita 1.4 (Gnome 45). It would make sense to use it here.

caffeine@patapon.info/preferences/timerPage.js
```diff
@@ -0,0 +1,346 @@
...
+        this.add(customDurationGroup);
+
+        // Init
+        this._activeCustomvalue();
+        this._updateTimerDuration(durationIndex);
+        this._updateResetButtonState();
+
+        // Bind signals
+        // --------------
+        enableCustomTimerSwitch.connect('notify::active', (widget) => {
```
gabrbrand, line 150:
```suggestion
        this.enableCustomTimerRow.connect('notify::active', (widget) => {
```
See above

gabrbrand (changes requested):

gabrbrand:
I manually installed this branch with `make all` on Fedora Rawhide (Gnome 46).
The `stopwatch-symbolic` currently isn't visible in dark style:
image
I could fix this by moving `caffeine@patapon.info/icons/stopwatch-symbolic.svg` to `caffeine@patapon.info/icons/hicolor/scalable/actions` and changing
```javascript
Gio.icon_new_for_string(`${this._path}/icons/${TimerMenuIcon}.svg`); 
```
to
```javascript
Gio.icon_new_for_string(`${this._path}/icons/hicolor/scalable/actions/${TimerMenuIcon}.svg`);
```
in `caffeine@patapon.info/extension.js` (line 147)

pakaoraki:
Thanks for reporting, icon fixed.

caffeine@patapon.info/extension.js
```diff
@@ -183,27 +182,34 @@ class CaffeineToggle extends QuickSettings.QuickMenuToggle {
...
+        for (const [index, timer] of durationValues.entries()) {
             let label = null;
-            if (timer[0] === 0) {
+            if (timer === 0) {
                 label = _('Infinite');
             } else {
-                label = parseInt(timer[durationIndex]) + _(' minutes');
+                let hours = Math.floor(timer / 3600);
+                let minutes = Math.floor((timer % 3600) / 60);
+                if (parseInt(hours) !== 0) {
```
stuarthayhurst, line 197: If `hours` is already `Math.floor(...)`, why is `parseInt()` called on it? `hours` will already be an integer at this point, unless I misunderstood

Same for `minutes`

caffeine@patapon.info/extension.js
```diff
@@ -797,8 +805,16 @@ class Caffeine extends QuickSettings.SystemIndicator {
     _updateTimerSubtitle() {
         if (!this._settings.get_boolean(TOGGLE_STATE_KEY)) {
             const timerDuration = this._settings.get_int(TIMER_KEY);
+            const hours = Math.floor(timerDuration / 3600);
+            const min = Math.floor((timerDuration % 3600) / 60);
+            let timeLabel = hours !== 0
+                ? parseInt(hours) + _(' hours ')
```
stuarthayhurst, line 811: Same here

caffeine@patapon.info/extension.js
pakaoraki: Absolutely, I'll fix it.

stuarthayhurst:
The pipeline failure seems to be eslint 9 complaining about arguments, we wouldn't need to try and fix it if we just drop the jsdoc stuff, since I don't think we don't really use it anyway

If we drop it we just need to update the makefile, workflow definition and linter configs

pakaoraki:
I tried to remove the jsdoc plugin but it's seams that we need to rewrite the entire Eslint config for version 9.0 ([migrate-to-9.0.0](https://eslint.org/docs/latest/use/migrate-to-9.0.0#flat-config)) as they changed the config file format.

Should we make a separate PR for this ?

stuarthayhurst:
> Should we make a separate PR for this ?

Yeah, probably a good idea

matdave:
This feature is awesome. I've been wanting long timers since I work on dual computers and need both on during my work day. This is saving me from having to remember to turn caffeine off at the end of my work day.

pakaoraki:
@eonpatapon @stuarthayhurst 
I think it's ready to merge, let me know if it's ok for you.

stuarthayhurst:
Looks ok to me, do we want to merge this with a squash merge since there are a lot of small commits and fixes for previous commits?

caffeine@patapon.info/preferences/timerPage.js
```diff
@@ -0,0 +1,341 @@
...
+            draw_value: false,
+            orientation: 'horizontal',
+            digits: 0,
+            adjustment: adjustSliderTimer
+        });
+        this.sliderTimer.add_mark(0, Gtk.PositionType.BOTTOM, null);
+        this.sliderTimer.add_mark(1, Gtk.PositionType.BOTTOM, null);
+        this.sliderTimer.add_mark(2, Gtk.PositionType.BOTTOM, null);
+        this.sliderTimer.add_mark(3, Gtk.PositionType.BOTTOM, null);
+        this.sliderTimer.add_mark(4, Gtk.PositionType.BOTTOM, null);
```
stuarthayhurst, line 83: Should this be a loop instead?

pakaoraki: Yes, I changed it.

stuarthayhurst:
Just a heads up, the locales don't need to be updated with every change if it's just line numbers that changed, it also makes reverting commits harder

I can't see anything else to change, LGTM. I'll rebase #334 when this is merged.

stuarthayhurst (approved):

stuarthayhurst (changes requested):
image

This doesn't seem right, hours shouldn't be plural here

caffeine@patapon.info/preferences/timerPage.js
```diff
@@ -0,0 +1,339 @@
...
+        if (shortValue !== parseInt(TIMERS_DURATION[durationIndex][0]) * 60) {
+            isCustomed = true;
+        }
+        if (mediumValue !== parseInt(TIMERS_DURATION[durationIndex][1]) * 60) {
+            isCustomed = true;
+        }
+        if (longValue !== parseInt(TIMERS_DURATION[durationIndex][2]) * 60) {
+            isCustomed = true;
+        }
+        return isCustomed;
```
stuarthayhurst, line 203:
```suggestion
        const selectors = [this.shortTimerSelector, this.mediumTimerSelector, this.longTimerSelector];
        const durationIndex = this.sliderTimer.get_value();
        let isCustom = false;

        selectors.forEach((selector, i) => {
            if (selector.get_value() !== parseInt(TIMERS_DURATION[durationIndex][i]) * 60) {
                isCustom = true;
            }
        });

        return isCustom;
```

Untested, but would something like this be better? I wasn't going to nit-pick this because it's not huge, but if you've already got to fix the other issue this might be worth looking at too.

pakaoraki: I will.

pakaoraki:
> image
> 
> This doesn't seem right, hours shouldn't be plural here (also applies when set to 1 minute).

I fix it: it will display properly the plural only when it should. I also improve the code as you suggest. Let me know if you see somethings else.

pakaoraki:
> Just a heads up, the locales don't need to be updated with every change if it's just line numbers that changed, it also makes reverting commits harder
> 
> I can't see anything else to change, LGTM. I'll rebase #334 when this is merged.

Ok I will stop doing that, it's also make it easiest. Then, is translation reference update when installing the extension ?

stuarthayhurst:
> > Just a heads up, the locales don't need to be updated with every change if it's just line numbers that changed, it also makes reverting commits harder
> > I can't see anything else to change, LGTM. I'll rebase #334 when this is merged.
> 
> Ok I will stop doing that, it's also make it easiest. Then, is translation reference update when installing the extension ?

Whoops, I missed the lines that did actually change in those commits, sorry. For my own extensions, I update the translations when a feature is completely ready since I find that simpler.

stuarthayhurst:
I'll run this for a couple days and let you know if anything goes wrong, otherwise it's looking good :)

stuarthayhurst (approved):

foolishgrunt:
Super stoked to see this feature added! Any idea when we'll see a new release?

stuarthayhurst:
> Super stoked to see this feature added! Any idea when we'll see a new release?

Most likely with the release to add GNOME 47 support, which should be around September time. In the meantime, you could always build from source, you'll just have to increase the version number to match the current release, otherwise GNOME will 'update' it.

PR #327, opened 2024-04-29
Force Eslint version 8.57.0

pakaoraki:
Eslint 9 seams to need a complete config rewrite: I tried to migrate the config file to version 9, but I'm not very familiar with it.
In the meantime, we can use the last stable version of Eslint 8 to keep process working.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -16,7 +16,7 @@ jobs:
         version: 1.0
     - name: Install eslint
       run: |
-        npm install -g eslint eslint-plugin-jsdoc
+        npm install -g eslint@8.57.0 eslint-plugin-jsdoc
 
     - name: Run linter target
       run: |


stuarthayhurst:
I'm guessing it's not worth using a package lock file for 2 packages? If that's the case, I agree with this, as it keeps things working for now :)

eonpatapon (approved):

matdave:
I work on many node-based projects, and would argue that the lack of a package / package lock file is the exact cause of this issue. Adding one to the repo would help with signalling a unified version to test against and run locally.

stuarthayhurst:
> I work on many node-based projects, and would argue that the lack of a package / package lock file is the exact cause of this issue. Adding one to the repo would help with signalling a unified version to test against and run locally.

I mean I don't see why we can't supply the files if it's worthwhile, but the issue seemed to be more that we didn't specify a version, not how we specified a version

pakaoraki:
I'm sorry because I'm not very familiar with workflow and I didn't really think about package lock file. If it's ok, we can maybe merge this as a temporary fix for the current PR and improve this later.

Issue #330, opened 2024-05-10
Feature Request: Triggers for automation


forabi:
Having the ability to control when caffeine is triggered similarly to the Mac app ["Amphetamine"](https://apps.apple.com/us/app/amphetamine/id937984704?mt=12) would be a great addition to this extension. The Mac app allows triggering the wake locks based on conditions like e.g. bluetooth headphone connected or based on a schedule (e.g. 9am-5pm), etc.

Issue #307, opened 2023-12-01
Support listening on D-BUS to MPRIS implementation to trigger caffeine


orontee:
# Motivation

VLC, Rhythmbox, Video, clementine, etc. there are many applications implementing the D-BUS interface Media Player Remote Interfacing Specification. From my pov, they're exactly the applications that should trigger Caffeine.

It would be nice to have an option for Caffeine to listen to D-BUS to get enabled when an application advertise it's playing through its MPRIS interface.

Gain for users: They would not have to register application one by one any more!

orontee:
Right, sorry for that.

Issue #217, opened 2022-11-06
Enable when playback is detected via MPRIS


carlocastoldi:
MPRIS is a D-Bus API which aims to provide a common programmatic API for controlling media players.
In our case, it is interesting because it provides a mechanism for querying compliant media players, including music players and browsers.

Using the MPRIS API to determine if there currently is any media playing sounds like good option for this extension!

eonpatapon:
Not sure to understand what you are proposing ?

I don't see how MPRIS/music players would interact with Caffeine

eonpatapon:
Ah you mean maybe an option to enable Caffeine when some player is 'play' status ?

carlocastoldi:
>Ah you mean maybe an option to enable Caffeine when some player is 'play' status ?
yes, correct! Similar to the option to enable caffeine for fullscreen apps

snoopdouglas:
This feature would be incredibly useful to me, given that I use HDMI audio - so screen off = audio off too.

I've never done any work on GNOME extensions but would be happy to start work on a PR for this. Any pointers would be appreciated.

eonpatapon:
Excerpts from snoopdouglas's message of February 13, 2024 9:49 am:
> This feature would be incredibly useful to me, given that I use HDMI audio - so screen off = audio off too.
> 
> I've never done any work on GNOME extensions but would be happy to start work on a PR for this. Any pointers would be appreciated.
> 

MPRIS status is exposed via dbus, so the extension would need to track via DBUS if some player is in playing state.

I think you can look for media player extensions and see how they use the MPRIS Dbus API.

Not the simplest change to do, will add quite some code to the extension.

carlocastoldi:
I implemented MPRIS trigger in [`caffeine-ng`](https://codeberg.org/WhyNotHugo/caffeine-ng). You might want to take a hint on the dbus calls I used. It can be useful: https://codeberg.org/WhyNotHugo/caffeine-ng/src/commit/1c638ba251124adb7e4812b717735f4c35fc3784/caffeine/triggers.py#L248

The PRs are [!127](https://codeberg.org/WhyNotHugo/caffeine-ng/pulls/129) and [!146](https://codeberg.org/WhyNotHugo/caffeine-ng/pulls/146)

PR #334, opened 2024-06-09
Refactor extension to use an inhibitor manager

### LINKED ISSUE #332: Freezes gnome shell when compiling unity project
When this extension is enabled. And I build an android APK on unity, gnome shell freezes for a few seconds to a few minutes. I have verified that this extension is causing it by individually enabling and disabling all extensions 1 for 1. Caffeine does not need to be poured for it to freeze, just enabling it causes this. 

If you need any logs of extra information, please let me know :)

stuarthayhurst:
This is a fairly extensive rewrite of `extension.js`, and I'm not proud of putting it in 1 commit, but absolutely nothing would work in the slightest until everything was moved over to an inhibitor manager, as I suggested in #285.

The current solution seems to create an inhibitor for every excuse to inhibit, which has lead to the extension losing track of its inhibitors in the past, and suffered from race conditions too. The new solution abstracts the inhibiting away to a new class, `InhibitorManager`. It connects to signals and calls `_updateState()`, which uses `_getInhibitReasons()` to scan every excuse to inhibit (user toggle, apps, fullscreen). It then compares these against the user's settings, and decides whether to inhibit or not, and updates the inhibitor and night light accordingly. This means we only every have 1 inhibitor, so we can't easily lose track of it. It also makes the app handling simpler, as we just add methods to handle the chosen app trigger, and return whichever app is responsible for triggering Caffeine.

This means adding new ways to enable and disable Caffeine just means making `_getInhibitReasons()` aware of them, making issues like #217, #307 and #330 much easier to approach.

In summary:
 - Slims down the codebase
 - Slims down DBus interfaces and GSettings schemas
 - Self-contains a lot of the code
 - Decouple the inhibitor logic and indicator logic
 - Creates the inhibitor synchronously, to put an end to the race conditions
 - Fixes #332 (I assume the old code got stuck in a loop of events)
 - Fixes a few miscellaneous bugs around triggering on workspaces

Any feedback and testing is appreciated :)

Closes #332

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -331,11 +331,14 @@ const InhibitorManager = GObject.registerClass({
                 this._isWaitingInhibit = false;
 
                 // Delete the inhibitor if we're not supposed to have it
-                if (!this._skipInhibit) {
+                /* We shouldn't need to check for this._isInhibited, but we
+                   might be able to recover from a bug if we do
+                */
+                if (this._skipInhibit || this._isInhibited) {
+                    this._removeInhibitor();
+                } else {
                     this._isInhibited = true;
                     this._inhibitorCookie = cookie;
-                } else {
-                    this._removeInhibitor();
                 }
             }
         );
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -125,6 +125,11 @@ const InhibitorManager = GObject.registerClass({
         this._lastReasons = [];
         this._ignoredReasons = [];
 
+        // App trigger signal IDs
+        this._appStateSignal = null;
+        this._focusWindowSignal = null;
+        this._workspaceSignal = null;
+
         // DBus proxies
         this._sessionManager = new DBusSessionManagerProxy(Gio.DBus.session,
             'org.gnome.SessionManager',
@@ -154,18 +159,54 @@ const InhibitorManager = GObject.registerClass({
             `changed::${INHIBIT_APPS_KEY}`,
             () => this._updateState(),
             `changed::${TRIGGER_APPS_MODE}`,
-            () => this._updateState(), this);
+            () => {
+                this._disconnectTriggerSignals();
+                this._connectTriggerSignals();
+                this._updateState();
+            }, this);
 
         // Update state when fullscreened
         global.display.connectObject('in-fullscreen-changed', () => this._updateState(), this);
 
         // Update when possible app triggers change
-        this._appSystem.connectObject('app-state-changed', () => this._updateState(), this);
-        global.display.connectObject('notify::focus-window', () => this._updateState(), this);
-        global.workspace_manager.connectObject('workspace-switched',
-            () => this._updateState(), this);
+        this._connectTriggerSignals();
 
         this._updateState();
+    }
+
+    _connectTriggerSignals() {
+        // Only connect to relevant signals for the selected trigger
+        switch (this._settings.get_enum(TRIGGER_APPS_MODE)) {
+        case AppsTrigger.ON_RUNNING:
+            this._appStateSignal = this._appSystem.connect('app-state-changed',
+                () => this._updateState());
+            break;
+        case AppsTrigger.ON_FOCUS:
+            this._focusWindowSignal = global.display.connect('notify::focus-window',
+                () => this._updateState());
+            break;
+        case AppsTrigger.ON_ACTIVE_WORKSPACE:
+            this._appStateSignal = this._appSystem.connect('app-state-changed',
+                () => this._updateState());
+            this._workspaceSignal = global.workspace_manager.connect('workspace-switched',
+                () => this._updateState());
+            break;
+        }
+    }
+
+    _disconnectTriggerSignals() {
+        if (this._appStateSignal !== null) {
+            this._appSystem.disconnect(this._appStateSignal);
+            this._appStateSignal = null;
+        }
+        if (this._focusWindowSignal !== null) {
+            global.display.disconnect(this._focusWindowSignal);
+            this._focusWindowSignal = null;
+        }
+        if (this._workspaceSignal !== null) {
+            global.workspace_manager.disconnect();
+            this._workspaceSignal = null;
+        }
     }
 
     _findRunningApp() {
@@ -404,11 +445,9 @@ const InhibitorManager = GObject.registerClass({
     }
 
     destroy() {
+        this._disconnectTriggerSignals();
         global.display.disconnectObject(this);
         this._settings.disconnectObject(this);
-        this._appSystem.disconnectObject(this);
-        global.display.disconnectObject(this);
-        global.workspace_manager.disconnectObject(this);
 
         if (this._isInhibited) {
             this._removeInhibitor();
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -666,6 +666,7 @@ class Caffeine extends QuickSettings.SystemIndicator {
             this._removeTimer();
         }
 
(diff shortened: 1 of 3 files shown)


stuarthayhurst:
Finished my testing, nearly everything works, except the night light control is completely unresponsive and the extension quickly re-enables itself if you disable it with a trigger active, but I still need to compare that to the old behaviour.

stuarthayhurst:
Fixed the night light, any issues with this are new to me

stuarthayhurst:
Also, this removes the `toggle-state` settings key to control its state, since it didn't really fit into the new structure. That does mean we lose CLI control, but I'm not sure people actually used this as there are other desktop apps that would be better suited to it, and it kind of seemed like a side-effect of the old design.

stuarthayhurst:
Apparently this closes #332 as well

stuarthayhurst:
@pakaoraki @eonpatapon GNOME 47.alpha is out, I'll start work on porting soon, but ideally this would be merged by then. Any chance of a quick review?

Thanks :)

pakaoraki:
Sorry for the late answer. I think it's a really nice improvement, I like this new approach. For now, this version is not working on my computer (Ubuntu 24/Fedora 40, Gnome 46): I have this error message "TypeError: app is null". I will test/debug it soon.

stuarthayhurst:
Do you have your app trigger set to active workspace only? I think I missed a check there.

pakaoraki:
I ran this version for the last few days.
I find some issues: 
- Apps trigger not working: it create multiple inhibitors again. I reckon, you removed some mechanism I have previously added. For some reasons, apps events are not clean and can be erratic, calling multiple time `_updateState()` function in a row (for few millisecond). The concept of having one inhibitor flag to manage is really great but I don't think it can handle all the race condition problem due to this events.

- There is a minor problem with the subtitle label when a timer have been selected: when caffeine is off, the duration is displayed as subtitle: for some reason here it's not displayed immediately, you need to switch workspace (or sometimes close and reopen the toggle menu) to refresh the label.

Also, if it's possible, it could be nice to keep the `toggle-state` settings key, but from what I saw, it's seams a bit tricky to add it.

caffeine@patapon.info/extension.js
```diff
@@ -127,12 +106,302 @@ const AppsTrigger = {
...
+
+        // Update inhibitor if required
+        if (this._isInhibited !== shouldInhibit) {
+            if (shouldInhibit) {
+                this._addInhibitor(reasons);
+            } else {
+                this._removeInhibitor();
+            }
+        }
+
```
pakaoraki, line 287: As it have only one inhibitor to manage, I recommend to destroy the previous inhibitor every time before creating the new one.  

I suggest to maybe store the cookie in a key (if it's possible) to make sure to remove it after the extension is badly stopped (crash). This can also be used to keep the inhibitor running after Caffeine been disabled/enabled when for example the screen is locked (it happened when using blank screen option).

stuarthayhurst: Mmm, I wondered about doing that, but I left it for now since it *shouldn't* be possible to incorrectly call them, and if they're being called at the wrong time we have a bug anyway.

I'll sort the current issues, then better protect the inhibitor for extra safety.

stuarthayhurst:
> * Apps trigger not working: it create multiple inhibitors again. I reckon, you removed some mechanism I have previously added. For some reasons, apps events are not clean and can be erratic, calling multiple time `_updateState()` function in a row (for few millisecond). The concept of having one inhibitor flag to manage is really great but I don't think it can handle all the race condition problem due to this events.

I ran into this during development but thought I fixed this, I'll take another look, thanks

> * There is a minor problem with the subtitle label when a timer have been selected: when caffeine is off, the duration is displayed as subtitle: for some reason here it's not displayed immediately, you need to switch workspace (or sometimes close and reopen the toggle menu) to refresh the label.

I've got a fix for that, thanks

> Also, if it's possible, it could be nice to keep the `toggle-state` settings key, but from what I saw, it's seams a bit tricky to add it.

It shouldn't be too hard to implement it back, I dropped it since it seemed like an artefact of the previous design and less of an intended feature, it allowed slimming down the gsettings schema and it was one less entry point to deal with. But if people miss it it can be added back, I'll tackle that after the new system works properly.

stuarthayhurst:
Since JavaScript is single-threaded, we can just test if we're waiting to create an inhibitor, as well as if we've already created one. If we need to delete and inhibitor that's waiting to be created, we can check for this in the callback to DBus, and just destroy it straight away.

This hasn't broken anything on my system, but I also don't know if it fixed the issue, since I couldn't reproduce it.

stuarthayhurst:
Hopefully that should be any actual bugs fixed. The trigger signals could be improved to connect based off of what it actually needs, and I'll look for an elegant way to bring back `toggle-state`.

caffeine@patapon.info/extension.js
```diff
@@ -127,12 +106,302 @@ const AppsTrigger = {
...
+        // Update state when fullscreened
+        global.display.connectObject('in-fullscreen-changed', () => this._updateState(), this);
+
+        // Update when possible app triggers change
+        this._appSystem.connectObject('app-state-changed', () => this._updateState(), this);
+        global.display.connectObject('notify::focus-window', () => this._updateState(), this);
+        global.workspace_manager.connectObject('workspace-switched',
+            () => this._updateState(), this);
+
```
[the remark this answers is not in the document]
stuarthayhurst: Implemented relevant signals suggestion. I'm not convinced on the signal blocker, since if we have an event happen and then revert in a short space of time, the revert would be missed and it would be stuck in the old state. I think that's just going to cause bugs down the line.

Filtering the app id feels a little redundant, as it would be trying to do part of `_updateState()`'s job for it, I think it's better to leave it as a sort of sync method, that just takes in the environment around it when given a nudge.

I'm happy that this point is resolved, if you're not just let me know :)

stuarthayhurst:
Additionally, this fixes a bug that moving an app between workspaces would sometimes be missed depending on which app was focused. I don't think that had been reported yet.

stuarthayhurst:
@pakaoraki Rebased since it's been a while, what's your position on this? I haven't had any issues on my system, but then again you found some bugs before that I couldn't reproduce.

stuarthayhurst:
Found a way to call the `Inhibit` method synchronously, so now the extension only has to manage a single inhibitor, synchronously. That should put an end to the race conditions, and also drops all the ugly code to try and work around it.

Let me know if anyone still has any issues with anything :)

pakaoraki:
Sorry, I've been really busy since weeks now. I will try again the code and review it again as soon as I can (before the weekend). Please test it fully apps trigger mode with different scenarios (open multiple apps, change virtual desktop, etc...) last time, multiple flags were created.

stuarthayhurst:
> Sorry, I've been really busy since weeks now. I will try again the code and review it again as soon as I can (before the weekend).

Yeah no worries, I've only just found time to pick it back up anyway

> Please test it fully apps trigger mode with different scenarios (open multiple apps, change virtual desktop, etc...) last time, multiple flags were created.

I'm running it on my machine and I haven't found any issues so far, I'll try to be as comprehensive as possible. I can't really test the night light as easily though, since it crashes my machine (NVIDIA at it again...)

pakaoraki:
Tested on last stable Fedora and Ubuntu, it's working great, no more issues with multiple apps flags. I will keep testing today but everything seams fine so far, thanks for this amazing work ! 
I let @eonpatapon to review it maybe as it is a big rewrite.

stuarthayhurst:
This would be great to get merged before GNOME 48 is ready, then it would benefit from all the testing done on that version.

@eonpatapon Do you think you'll get time to review this? No worries if you can't, I know how busy everyone can be :)

PR #345, opened 2024-10-05
Update the Polish translation & bump to v55

actrawave:


stuarthayhurst:
Looks good, thanks :)

PR #347, opened 2024-11-01
Settings button in quicksettings

ChrisLauinger77:
fast way to access extension settings


ChrisLauinger77:
Bildschirmfoto vom 2024-11-01 07-57-37

stuarthayhurst:
If this is a feature we want, the parameters might need another look. If we're passing a reference to `CaffeineExtension` in,  then passing the settings, path and name in is redundant, since they can all be acquired from the `CaffeineExtension` reference

ChrisLauinger77:
Made the suggested changes

stuarthayhurst:
@eonpatapon @pakaoraki Is this a feature we want? I'm happy with it since it's in line with the rest of the shell, but I don't want to merge it unless everyone else is happy with it

@ChrisLauinger77 Could you fix the merge conflicts please, and sort the linter issues? Other than that LGTM :)

ChrisLauinger77:
I will look into it in the evening

stuarthayhurst:
You should be able to see under the Checks tab -> lint -> Run lint target dropdown

ChrisLauinger77:
Now I can see it - it did not show up after I pushed and waiter 30 min

ChrisLauinger77:
I think I have it all now ...

stuarthayhurst (approved):
Working in my VM, LGTM

Also tested with #355, correctly hidden on the lock screen too

pakaoraki:
Sorry for the very long delay. I had the same idea some time ago, but I never implemented it. So I personally like this feature.

PR #349, opened 2024-11-07
Fix hungarian translation

mobal:
The Hungarian localization contains several errors, and with this PR, I would like to fix them.

Képernyőkép 2024-11-07 10-28-20

"percek" -> "perc"
"órák" -> "óra"
"órá" -> does not exist in the hungarian language, contains a typo, "óra"


stuarthayhurst:
Thanks :)

Issue #132, opened 2018-12-06
gets deactivated as soon screen gets locked


spi43984:
When I have a long running task and enable caffeine, I often lock my screen and go for a coffee ;-) But caffeine gets deactivated when I lock the screen.

kwilczynski:
Hi @spi43984,

There have been similar requests in the past to do this or to disable such behaviour, or to allow users to enable/disable state restoration as per:

https://github.com/eonpatapon/gnome-shell-extension-caffeine/issues/53
https://github.com/eonpatapon/gnome-shell-extension-caffeine/issues/56
https://github.com/eonpatapon/gnome-shell-extension-caffeine/issues/58

The current implementation with an option to remember state seems to be what things have settled at.

mlo-poc:
I notice, that this report is quite old but I still see the need.

I switched back to GNOME a few days ago and found that there is no simple way to **lock the screen** and keep the system **alive** (e.g. not going to suspend mode) at the same time.

this is quite important to me because I sometimes want my screen locked for security but still be able to connect remotely - which of course is not possible if the systems goes to suspend mode.

tutacat:
> I notice, that this report is quite old but I still see the need.
> 
> I switched back to GNOME a few days ago and found that there is no simple way to **lock the screen** and keep the system **alive** (e.g. not going to suspend mode) at the same time.
> 
> this is quite important to me because I sometimes want my screen locked for security but still be able to connect remotely - which of course is not possible if the systems goes to suspend mode.

You could disable auto sleep and use manual sleep

mlo-poc:
> You could disable auto sleep and use manual sleep

well ...
I surely could do that, but I usually **want** my computer to go to sleep when I leave it for a while. It's something I just don't think about in day to day operation. I use the Pause-Key for locking my screen on all my computers, so this is wired very deeply in my muscular memory and I love caffeine, because it is so convenient to have a one-click to disable the screensaver.

It used to work like that for ages: One click to activate caffeine (or similar) + press Pause to enable screensaver and computer is in *locked but don't sleep* state.

With your suggestion this state is at least six clicks away. Not very user friendly.

Janhouse:
Seems like this still has not been resolved. Is this not technically possible or is it just a matter of someone adding that feature?

eonpatapon:
Well the issue was that the lock screen disables extensions, and when Caffeine gets disabled all inhibitors are removed.

[...]

PR #355, opened 2025-01-06
Don't disable the extension in the lock screen

eonpatapon:
Ref #132


stuarthayhurst:
Working on my systems, doesn't interfere with suspending my laptop either

PR #360, opened 2025-02-13
Feature Mpris Media Player2 dbus support

### LINKED ISSUE #217: Enable when playback is detected via MPRIS

PVermeer:
Closes #217

#### Mpris implementation
I've been wanting to learn some GTK en dbus coding so I went on and implement this feature.

I've created a singleton class that should handle all the dbus mpris communication in a module (file). This should keep things simple. It exposes the signal `isPlaying` that emits `true` or `false` if an mpris player is playing. In the future it could expose more data. It also exposes a manual `refresh()` and the `isPlaying` property on the instance.

Class should be used with the static members `Get()` and `Destroy()`. On the instance connecting with `connect('isPlaying', (_, isPlaying) => {})` should pass the playing status.

I have a TypeScript background so I like types and have added some useful types via JsDocs. I also added vscode dev-container files for easy development on vscode. These can be removed of course. 

I've tested this on my systems and it seems to work fine.

Let me know what you think.


Makefile
```diff
@@ -10,7 +10,8 @@ build:
 	cd $(EXTENSION_DIR); \
 	gnome-extensions pack --force --podir=locale \
 	                      --extra-source=preferences/ \
-	                      --extra-source=icons/; \
+	                      --extra-source=icons/ \
+												--extra-source=mprisMediaPlayer2.js; \
```
stuarthayhurst, line 14: Is there a reason this is so heavily indented?

.devcontainer/devcontainer.json
stuarthayhurst, line 1: I don't personally use VSCode, so I can't really maintain or test these, unless another maintainer is willing to, I'm leaning towards dropping them

caffeine@patapon.info/mprisMediaPlayer2.js
stuarthayhurst, line 1: Is there a reason the method names start with `#`?

Makefile
PVermeer: It's not in my ide. I see what happend. My line is all tabs. Other lines are tab + a bunch of spaces. Will fix.

.devcontainer/devcontainer.json
PVermeer: Dropped

caffeine@patapon.info/mprisMediaPlayer2.js
PVermeer: Yes, this is in the ECMA standard as a replay to TypeScript's `private` syntax. These are marked private and cannot be accessed outside of classes (creates an error). It basically replaces `_` with a real internal (private) syntax.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

PVermeer:
> Thanks for working on this, but I think this would also need to be put behind a setting

Yes. I was not sure if it should be a core feature or a setting. I went with core feature. I think it would be best if this a setting that is enabled by default?

Let me know.

stuarthayhurst:
> I think it would be best if this a setting that is enabled by default?

I'm not sure about that, since we haven't had many requests for this and it's not the main focus of the extension

I'm open to other opinions, but since this is a change in behaviour from previous versions, I think it should be opt-in.

caffeine@patapon.info/mprisMediaPlayer2.js
stuarthayhurst: Ah I wasn't aware of that, thanks.

Personally, I'd go with `_` since that's what the rest of the code base uses, and it's the convention for GNOME Shell's upstream JavaScript, which makes sense to mirror when our code is tightly coupled to that.

PVermeer: Makes sense. I changed it.

eonpatapon:
I also think it should be opt-in

@PVermeer also big changes have landed on master, so will need to adapt the code to integrate with the new InhibitorManager

PVermeer:
I think this is ready. I've implemented it in the new InhibitorManager the same as the fullscreen inhibitor and tested it on my systems.

It works great for me in Spotify and browser YouTube/Twitch where I don't have to add these to the apps feature anymore. My systems now go to sleep when nothing is playing and the apps are still open.

Let me know if you still need changes.

caffeine@patapon.info/schemas/org.gnome.shell.extensions.caffeine.gschema.xml
```diff
@@ -77,6 +77,11 @@
         <summary>Enable when a fullscreen application is running</summary>
         <description>Enable when a fullscreen application is running</description>
     </key>
+    <key type="b" name="enable-mpris">
+        <default>false</default>
+        <summary>Enable when an application is playing media</summary>
+        <description>Enable when an application is reporting playback status via mpris</description>
```
stuarthayhurst, line 83:
```suggestion
        <description>Enable when an application is reporting playback status via MPRIS</description>
```

caffeine@patapon.info/mprisMediaPlayer2.js
```diff
@@ -0,0 +1,342 @@
...
+ */
+const DBusMprisPlayerInterface = `<node>
+  <interface name="org.mpris.MediaPlayer2.Player">
+    <property name="PlaybackStatus" type="s" access="read"/>
+  </interface>
+</node>`;
+
+class PrivateContructorParams {
+    dont = 'doit';
+}
```
stuarthayhurst, line 67: What's this for?

caffeine@patapon.info/mprisMediaPlayer2.js
```diff
@@ -0,0 +1,342 @@
...
+    _onNameOwnerChanged(_proxy, _sender, [name, oldOwner, newOwner]) {
+        if (!name.startsWith(this._mprisPrefix)) {
+            return;
+        }
+        if (oldOwner && !newOwner) {
+            this._removePlayer(name);
+        }
+        if (newOwner && !oldOwner) {
+            this._addPlayer(name);
+        }
```
stuarthayhurst, line 252: Is something like this a bit clearer?

```suggestion
        if (oldOwner !== newOwner) {
            if (newOwner) {
                this._addPlayer(name);
            } else if (oldOwner) {
                this._removePlayer(name);
            }
        }
```

If they're guaranteed that both are never set at the same time, `if (oldOwner !== newOwner)` could also be dropped in that

caffeine@patapon.info/mprisMediaPlayer2.js
```diff
@@ -0,0 +1,342 @@
...
+
+        if (!(params instanceof PrivateContructorParams)) {
+            throw new TypeError(
+                'MprisMediaPlayer2 is not constructable. ' +
+                'Use `MprisMediaPlayer2.Get()` and `MprisMediaPlayer2.Destroy()`'
+            );
+        }
+
+        this._DBusProxy = Gio.DBusProxy.makeProxyWrapper(DBusInterface);
```
stuarthayhurst, line 305: Can this be local to this function? It doesn't seem to be referenced outside of it

caffeine@patapon.info/mprisMediaPlayer2.js
PVermeer: I don't know. Tbh I could not find clear documentation for this. All I know is that code, as is, works and is tested.

 `oldOwner !== newOwner` can these be compared? 
`If they're guaranteed that both are never set at the same time` don't know if that' s guaranteed.

All I found was that if there is an `oldOwner` and no `newOwner` that means an app is removed and visa versa. So I implemented and tested this.

stuarthayhurst: Found this https://dbus.freedesktop.org/doc/dbus-specification.html#bus-messages-name-owner-changed

Seems that if `newOwner` isn't set then it was removed and if `oldOwner` isn't set it was added

```suggestion
    if (newOwner === "") {
        this._removePlayer(name);
    } else if (oldOwner === "") {
        this._addPlayer(name);
    }
```

Completely untested, I don't know if they get set to `undefined`, `null` or `""` but in principle, that should work

caffeine@patapon.info/mprisMediaPlayer2.js
PVermeer: Private constructor logic.

https://github.com/PVermeer/gnome-shell-extension-caffeine/blob/b1ec5df0c509f75d9a56a44ed484260a195fc4c1/caffeine%40patapon.info/mprisMediaPlayer2.js#L298-L303

I implemented MPRIS as a singleton service/plugin to this extension. To avoid leaks I made it a singleton class that should be manually constructed. It's just an object thats only available in this module so the class can only be constructed via static methods.

stuarthayhurst: Is this necessary since the class won't be accessed outside of Caffeine?

Personally, I don't think the class needs to be as defensive, since it's documented that `Get()` and `Destroy()` should be used:

https://github.com/PVermeer/gnome-shell-extension-caffeine/blob/b1ec5df0c509f75d9a56a44ed484260a195fc4c1/caffeine%40patapon.info/mprisMediaPlayer2.js#L328

caffeine@patapon.info/mprisMediaPlayer2.js
PVermeer: It could be. It's a dependency of this class that methods could use if needed... This is just standard initialization of the class.

caffeine@patapon.info/mprisMediaPlayer2.js
PVermeer: It's just a safety measure. In other languages one can simple define the constructor private. It doesn't hurt or  effect anything besides accidental class instantiating in future work.

You can change it if you like ofc

caffeine@patapon.info/mprisMediaPlayer2.js
PVermeer: I tested your suggestion following the dbus spec and commited the change.

caffeine@patapon.info/mprisMediaPlayer2.js
PVermeer: It also add the `@typedef` on the property.

https://github.com/PVermeer/gnome-shell-extension-caffeine/blob/afec6f609f215aa82cfa325eee899be8f917a829/caffeine%40patapon.info/mprisMediaPlayer2.js#L20-L29

I recommend to keep it.

caffeine@patapon.info/mprisMediaPlayer2.js
PVermeer: I removed this logic

caffeine@patapon.info/mprisMediaPlayer2.js
```diff
@@ -0,0 +1,322 @@
...
+        // Declare some GObject.Object properties (non typescript has some difficulty with this weird import)
+        if (!this.emit) {
+            this.emit = super.emit;
+        }
+        if (!this.connect) {
+            this.connect = super.connect;
+        }
+        if (!this.disconnect) {
+            this.disconnect = super.disconnect;
+        }
```
stuarthayhurst, line 282: This seems odd to me, is this because the class isn't registered with `GObject.registerClass`?

PVermeer: I removed this. This was for type checking failures. Serves no purpose anymore.

caffeine@patapon.info/mprisMediaPlayer2.js
PVermeer: I removed it anyways.

stuarthayhurst:
Working in my VM, thanks for working on this :)

stuarthayhurst (approved):

PhrozenByte:
I tested this together with #362 with GNOME 47, works like a charm :rocket:

I'm running this in production now, so if any issues emerge I'll report them, else consider everything working for me.

> I'm not sure about that, since we haven't had many requests for this and it's not the main focus of the extension

I'm naturally just speaking for myself here, but that covers about 90% of the reasons I used to enable Caffeine manually for - and I didn't even make the connection with media players before seeing this PR :laughing: Utilizing the MPRIS API really is an excellent idea @PVermeer, thanks for your work! :+1:

caffeine@patapon.info/extension.js
```diff
@@ -168,6 +172,9 @@ const InhibitorManager = GObject.registerClass({
         // Update state when fullscreened
         global.display.connectObject('in-fullscreen-changed', () => this._updateState(), this);
 
+        // Update state when mpris Player Playback status changed
+        MprisPlayer.Get().connectIsPlaying((_isPlaying) => this._updateState());
```
eonpatapon, line 176: Shouldn't we listen on dbus only if the mpris setting is enabled?

PVermeer: Yes we should. I did not implement this because the InhibitorManager did not support this and this is isn't done anywhere else (like for fullscreen). Untested but I think the performance impact would be negligible.

My suggestion would be:

```suggestion
        // Update state when extension settings changed
        this._settings.connectObject(
            `changed::${SCREEN_BLANK}`,
            () => this._forceUpdate(),
            `changed::${FULLSCREEN_KEY}`,
            () => this._updateState(),
            `changed::${MPRIS_KEY}`,
            () => this._handle_mpris_setting(),
            `changed::${NIGHT_LIGHT_KEY}`,
            () => this._updateState(),
            `changed::${INHIBIT_APPS_KEY}`,
            () => this._updateState(),
            `changed::${TRIGGER_APPS_MODE}`,
            () => {
                this._disconnectTriggerSignals();
                this._connectTriggerSignals();
                this._updateState();
            }, this);

        // Update state when fullscreened
        global.display.connectObject('in-fullscreen-changed', () => this._updateState(), this);

        // Init mpris
        this._handle_mpris_setting();
```
And:

```suggestion
    _handle_mpris_setting() {
        const enable = this._settings.get_boolean(MPRIS_KEY);
        if (enable && !MprisPlayer.isActive) {
            MprisPlayer.Get().connectIsPlaying((_isPlaying) => this._updateState());
        } else {
            MprisPlayer.Destroy();
        }
        this._updateState();
    }

```

PVermeer: Implemented it in e3234f6db4963f1628035d41610e78b1cd5226c2

I tested it, seems to work well. Let me know if it's alright.

PVermeer:
Can this be merged?

eonpatapon:
Yes, can you rebase and squash commits ?

PVermeer:
I've updated the PR with master.

You can merge however you like:
alt text

eonpatapon:
Thank you!

PR #362, opened 2025-02-16
Change default inhibit flags

### LINKED ISSUE #353: Manual suspend doesn't work anymore.
After updating my system (Arch Linux) a few days ago, manually suspending has stopped working. The button in the control center does nothing, and the power button prompts me to shut down the PC.

Running `systemctl hybrid-sleep` returns this:
`Operation inhibited by "username" (PID 1277 "gnome-session-b", user username), reason is "user session inhibited".`

I tried disabling my extensions one by one, and the one causing this is Caffeine. Also, if I just disable Caffeine from the control center, that allows me to suspend too.

stuarthayhurst:
Some systems seem to be having problems with manual suspend and shutdown with Caffeine enabled

Reportedly, using `8` as the inhibit flag instead of `12` fixes this. This means using `4` to disable suspend but allow the screen to blank will still have issues, so add a warning to the settings page.

Marking as draft so this can be tested properly, as it seems very configuration dependent

Closes #353


stuarthayhurst:
It seems to be working on my systems, but the issue it's supposed to fix only showed up around once a month for me...

Glad it's working for you, hopefully some more users are willing to test it and feedback :)

limblendokin:
Just tested it and it works for me too

Lawstorant:
Doesn't seem like it has any adverse effects on other functionality and this allows me to put my device to sleep when using steam big picture mode.

Another thing this fixed for me is that during work, i switch between two accounts and even if just one had caffeine enabled, I couldn't put my laptop to sleep. Changing the inhibit flag fixed this issue.

Is there anything the prevents this from being merged right now?

stuarthayhurst:
Sure, I haven't had any problems with this so I'll mark it ready

stuarthayhurst:
@eonpatapon @pakaoraki It seems a lot of users have issues that this fixes, have either of you tested it? I'm hesitant to merge a change to the inhibit flags, but the current ones seem broken for enough people anyway.

PhrozenByte:
@stuarthayhurst How can we help you to get this merged? If it's just testing: I run this in production for about half a year now. It doesn't just work in practice, but it's also the actually correct solution from a theoretical / tech docs standpoint. So, I'm pretty sure it's safe to merge :+1: Even in some weird freak accident situation that it somehow breaks Caffeine for many other users, it shouldn't be a big deal to simply revert it anyway. Thanks :heart:

PhrozenByte:
Yeah, okay, but why? :confused:

Maybe I miss something here, but it makes absolutely no difference in practice, does it? Caffeine creating an idle+suspend inhibitor (flags `12`) has the exact same practical effect as just creating an idle inhibitor (flags `8`), because the system can only auto-suspend if it's idle. The idle inhibitor already prevents the system from going idle, i.e. it won't ever auto-suspend, it doesn't need a suspend inhibitor for that.

Using the "Allow blank" option is a different story though, because then the system must be allowed to go idle, but not to auto-suspend. This indeed only works with an suspend inhibitor (flags `4`). That's why you didn't change anything when the "Allow blank" option is enabled and just added a note to the settings, right?

So, the only difference merging this PR makes is whether Caffeine is broken for users with systemd v257+ (which e.g. just landed in Debian Trixie) or not. Other users aren't affected at all. Thus I don't see any harm in simply merging it?

By the way, there isn't even a feature request in GNOME Session to support weak inhibitors (at least I searched for one just now, but didn't find any). That systemd v257+ changed the inhibitor logic is known to the GNOME devs though, they addressed that change in GNOME 48.0, but decided against implementing support for it with the DBus service.

If you really don't want to merge it yourself, can you please ping and/or request reviews from other maintainers? I'm not sure whether someone is monitoring this?

stuarthayhurst:
Because a weak inhibitor would exactly match how things were before systemd / GNOME updates broke things.

I wasn't in a hurry to get this merged because we normally do a release per GNOME version, unless something significant changes. With GNOME 49 coming up and the merge request for that having already landed, weak inhibitors won't be coming in time and most likely this'll be merged instead.

I'm happy to merge this myself, but generally I wait for another maintainer to review changes before merging them. In any case, only @eonpatapon can make a release, so merged or not this won't be making it to users before he makes a release.

PR #364, opened 2025-03-06
Bugfix/add missing show quick settings toggle hungarian translation

mobal:
I found two missing hungarian translations (missing from localisation file). With this PR I would like to fix it.


stuarthayhurst:
Could you also deal with the 2 translations marked as fuzzy please? If they're already correct you can just remove the fuzzy tag.

stuarthayhurst:
Thanks :)

PR #368, opened 2025-05-03
Fix Russian translation

FraxinusAcer:
Updated the Russian translation.

Files of other languages have been synchronized to the current version.

### DIFF
--- a/caffeine@patapon.info/locale/ru.po
+++ b/caffeine@patapon.info/locale/ru.po
@@ -1,79 +1,83 @@
+#
+# Fraxinus Acer <>, 2025.
+#
 msgid ""
 msgstr ""
 "Project-Id-Version: gnome-shell-extension-caffeine\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-11-01 07:51+0100\n"
-"PO-Revision-Date: 2023-09-11 21:41+0300\n"
-"Last-Translator: Aleksandr Melman <Alexmelman88@gmail.com>\n"
+"POT-Creation-Date: 2025-05-03 13:42+0300\n"
+"PO-Revision-Date: 2025-05-03 13:58+0300\n"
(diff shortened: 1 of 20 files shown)


caffeine@patapon.info/locale/ru.po
```diff
@@ -1,79 +1,83 @@
...
 msgstr " часы "
 
-#: extension.js:219
+#: extension.js:576
 msgid " minute"
 msgstr " минута"
 
-#: extension.js:222 preferences/timerPage.js:206
+#: extension.js:579 preferences/timerPage.js:206
+#, fuzzy
```
stuarthayhurst, line 52: Could you take care of the fuzzy translations for this file please? If they're fine you can just remove the tag :)

stuarthayhurst:
Some of the fuzzy translations in the other language files that were synchronised seem to be duplicates of other interface strings, so they're going to be mislabelled. It's probably better to leave those files as outdated so those strings are untranslated and then someone can come and translate them, or at the least we're not showing something _incorrect_.

caffeine@patapon.info/locale/ru.po
FraxinusAcer: After updating the translation file, I used this [ https://gitlab.gnome.org/GNOME/gtranslator/ ] program and made changes only to the untranslated lines.

I don't understand the code, and I don't know why the previous translator marked these lines like that. If I see these lines in the UI, I will fix fuzzy and improve them.

stuarthayhurst: It gets marked as fuzzy when the translations are updated from the code after something changes and it's not quite a match anymore. Instead of just erasing the string and leaving it untranslated, it marks it as fuzzy and leaves it there.

FraxinusAcer:
Thank you for your attention. 
I fixed these my mistake, but I was unable to change this pull request.
See: https://github.com/eonpatapon/gnome-shell-extension-caffeine/pull/369

PR #379, opened 2025-06-05
Fallback To Adawaita Style Icons 

ochi12:
image
image

I noticed that some of the fallback icons are Yaru Style so those icons Adawaita style. Theming is unaffected and I'm actually currently on PR in Yaru for the yaru themed icons


stuarthayhurst:
Should the apps icon be so blurry? Other than that, looks good :)

PR #383, opened 2025-06-09
Replace Depracated App Chooser with Custom One

ochi12:
Stated in gjs-docs
image

Custom Implementation Requires Gtk.Window or Adw.Window as Base Class.
Gtk TreeView is also depracted replaced by Gtk.ListView

### new app chooser as of this PR
image
<br>
made it as close as the original one.

### todo
- [x]  implement double click handling on single selection item 
- [ ]  handle Force Quit
- [x]  auto refresh app list when new app is installed or uninstalled
- [x] handled proper Tab Navigation
- [x]  handle scroll bug
- [x]  updated source to smaller chunnks
- [ ]  clean up source (eg. change all call backs to callername + CB as prefix)

### DIFF
--- a/caffeine@patapon.info/preferences/appsPage/appChooser.js
+++ b/caffeine@patapon.info/preferences/appsPage/appChooser.js
@@ -112,17 +112,19 @@ const CustomAppList = GObject.registerClass({
         this._query = null;
         this._desktopAppInfoCache = new Map();
 
-        const baseModel = this._generateBaseModel();
-        const filterModel = new Gtk.FilterListModel({
-            model: baseModel,
-            filter: Gtk.CustomFilter.new(this._filterCallBack.bind(this))
+        this._baseModel = this._generateBaseModel();
+        this._filter = Gtk.CustomFilter.new(this._filterCallBack.bind(this));
+        this._filterModel = new Gtk.FilterListModel({
+            model: this._baseModel,
+            filter: this._filter
         });
-        const sortModel = new Gtk.SortListModel({
-            model: filterModel,
-            sorter: Gtk.CustomSorter.new(this._sorterCallBack.bind(this))
+        this._sorter = Gtk.CustomSorter.new(this._sorterCallBack.bind(this));
+        this._sortModel = new Gtk.SortListModel({
+            model: this._filterModel,
+            sorter: this._sorter
         });
 
-        this._singleSelection = Gtk.SingleSelection.new(sortModel);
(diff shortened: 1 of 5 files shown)


stuarthayhurst:
> But what should be added is a way to refresh the list for new apps installed. There is actually a way to do it with Gio

There's a signal on `Shell.AppSystem` that you can use to detect changes to the installed apps if you weren't aware

ochi12:
> > But what should be added is a way to refresh the list for new apps installed. There is actually a way to do it with Gio
> 
> There's a signal on `Shell.AppSystem` that you can use to detect changes to the installed apps if you weren't aware
Oh yeah I meant Shell.AppSystem kinda mixed it up. Its been I while since updated my splash screen indicator extension

ochi12:
> > But what should be added is a way to refresh the list for new apps installed. There is actually a way to do it with Gio
> 
> There's a signal on `Shell.AppSystem` that you can use to detect changes to the installed apps if you weren't aware

If I implement this, should it be an auto refresh or refresh by button click?

stuarthayhurst:
> If I implement this, should it be an auto refresh or refresh by button click?

Auto-refresh probably makes more sense, it's not like it's a really active signal that's going to slow the interface down with hundreds of refreshes.

> Also since its a new change I might as well sleep in some search variation like matching

I'm not sure I follow, sorry

ochi12:
implemented better search
[Screencast From 2025-06-10 01-42-31.webm](https://github.com/user-attachments/assets/98270ddf-d4cc-4f30-bf6a-409b4dcc6414)

ochi12:
Implemented better search filter
[Screencast From 2025-06-10 01-42-31.webm](https://github.com/user-attachments/assets/45537e6d-377a-4bfb-8bcd-954c6786ef5f)

caffeine@patapon.info/preferences/appsPage/appChooser.js
```diff
@@ -0,0 +1,368 @@
...
+   version 3 of the License, or (at your option) any later version.
+
+   Caffeine is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+   without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+   See the GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License along with Caffeine.
+   If not, see <https://www.gnu.org/licenses/>.
+
+   Copyright 2022 Pakaoraki
```
stuarthayhurst, line 15: Is this correct?

PR #385, opened 2025-07-07
Add Ukranian translation

freeducks-debug:


caffeine@patapon.info/locale/uk.po
```diff
@@ -0,0 +1,417 @@
...
+
+#: preferences/generalPage.js:200 preferences/generalPage.js:232
+msgid "New accelerator…"
+msgstr "Нове скорочення…"
+
+#: preferences/timerPage.js:41
+msgid "Timer"
+msgstr "Таймер"
+
+# Не знаю як перекласти "preset".
```
stuarthayhurst, line 245: I don't know any Ukrainian I just used Google translate to sanity check the translations, did you need more context for preset? We can add a translators comment in the source code that gives more context for translators.

PR #403, opened 2025-09-25
Restore CLI

### LINKED ISSUE #370: CLI commands no longer working
I have been using the commands below for some time, but recently they have stopped working with the error message `No such key “toggle-state”`

```
# Check
gsettings --schemadir ~/.local/share/gnome-shell/extensions/caffeine@patapon.info/schemas/ get org.gnome.shell.extensions.caffeine toggle-state

# Enable
gsettings --schemadir ~/.local/share/gnome-shell/extensions/caffeine@patapon.info/schemas/ set org.gnome.shell.extensions.caffeine toggle-state true

# Disable
gsettings --schemadir ~/.local/share/gnome-shell/extensions/caffeine@patapon.info/schemas/ set org.gnome.shell.extensions.caffeine toggle-state false

```

The enable and disable commands are referenced in https://github.com/eonpatapon/gnome-shell-extension-caffeine/issues/257

If this key was intentionally removed, is there another way to change the state from the command line?

Extension version 57
Gnome Version 47
Ubuntu 24.10

stuarthayhurst:
Reintroduce the CLI with `cli-toggle` by ignoring updates when it matches the state and simulating a click when it doesn't.

Uses `cli-toggle` instead of `toggle-state` to avoid confusion, since GSettings keys aren't usually used as a CLI

Closes #370

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ This extension supports GNOME Shell `3.4` -> `49`
 | gnome-shell-3.10-3.30   |     - | GNOME 3.10 -> 3.30     |
 | gnome-shell-before-3.10 |     - | GNOME 3.4 -> 3.8       |
 
-## Installation From source
+## Installation from source
 
 ```bash
 make build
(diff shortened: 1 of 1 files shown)


BurntNail:
Not sure if I've just fucked up removing the master version or if there's an issue in the schema - the CLI commands don't work for me, and it looks like the schema in the installed schema directory doesn't have the `cli-toggle` key.

eonpatapon (approved):

PR #421, opened 2026-02-28
Fix lint workflow

stuarthayhurst:
The workflow was selecting the latest version of ESLint, which the config wasn't ready for, causing lint failures:
 - Fix the config for the newer versions
 - Update to the latest version
 - Tell the workflow to use the configured version

Should fix the failures in #419 and #420


stuarthayhurst:
Merging this so I can get some of the longer standing pull requests merged, and allow GNOME 50 support to go in smoothly

PR #430, opened 2026-05-20
Improved pt_BR translation to be more natural

Raffa064:
- Added missing translations
- Improved some translations to be more natural
- Fixed app's name gender to be gramatically correct


stuarthayhurst:
Thanks!

PR #432, opened 2026-07-13
Add option to prevent sleep when the laptop lid is closed

mattgia:
## Summary

This PR adds a new setting that allows Caffeine to prevent the laptop from going to sleep when you close the lid.
## How it works

- A new toggle, "Inhibit sleep on lid close", has been added to the General settings tab. It is disabled by default.
- When enabled, the extension now communicates directly with systemd-logind (the system service that handles power buttons and lid sensors). It requests a "block" on the lid-switch handler as long as Caffeine is active.

## Changes

1. Schema: Added inhibit-lid boolean key.
2. Preferences: Added a switch row in the General settings page.
3. Core Logic: Added a D-Bus proxy for org.freedesktop.login1 to manage the system-level inhibitor lock using Unix file descriptors.

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -57,6 +57,7 @@ const INDICATOR_POSITION = 'indicator-position';
 const INDICATOR_INDEX = 'indicator-position-index';
 const INDICATOR_POS_MAX = 'indicator-position-max';
 const CLI_TOGGLE_KEY = 'cli-toggle';
+const INHIBIT_LID_KEY = 'inhibit-lid';
 
 const ColorInterface = '<node> \
   <interface name="org.gnome.SettingsDaemon.Color"> \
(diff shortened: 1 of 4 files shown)


caffeine@patapon.info/extension.js
```diff
@@ -389,8 +415,8 @@ const InhibitorManager = GObject.registerClass({
             // Inhibit suspending the session or computer
             inhibitFlags = 4;
         } else {
-            // Inhibit the session being marked as idle
-            inhibitFlags = 8;
+            // Inhibit both suspending (to prevent lid-close sleep) and idle
+            inhibitFlags = 4 | 8;
```
stuarthayhurst: This was changed in #362 because it was causing problems, has anything else changed to fix that?

caffeine@patapon.info/extension.js
```diff
@@ -411,6 +437,36 @@ const InhibitorManager = GObject.registerClass({
...
+                                }
+                            }
+                        } catch (e) {
+                            log(`Caffeine: Failed to finish logind inhibitor: ${e.message}`);
+                        }
+                    });
+            } catch (e) {
+                log(`Caffeine: Failed to call logind inhibitor: ${e.message}`);
+            }
+        }
```
stuarthayhurst, line 464: I've only skim-read this, but is this asynchronous? Without special handling, we risk losing track of the inhibitor. Also, why do we need a second inhibitor? Did the change to the flags not cover the lid close? If this is necessary, why was the flag change necessary?

PR #433, opened 2026-07-15
perf: non-blocking DBus IPC calls and timer optimization

ducky-duke:
## Description
This PR addresses main-thread UI blocking bottlenecks and GLib timer issues:

1. **Async DBus Inhibit Calls**:
   - Replaced synchronous  in `extension.js` with async `InhibitRemote(...)`.
   - Added handling for in-flight requests if uninhibit is triggered prior to DBus callback response.

2. **Async MPRIS Player DBus Discovery**:
   - Replaced synchronous `ListNamesSync()` in `mprisMediaPlayer2.js` with async `ListNamesRemote(...)` to avoid stalling GNOME Shell on startup/refresh.

3. **Timer Logic Consolidation & Safe Source Removal**:
   - Consolidated dual GLib timeouts (`_timePrint` & `_timeOut`) into a single 1-second interval GLib source to prevent timer drift.
   - Guarded `GLib.Source.remove()` checks in `_removeTimer()` against null handles and already-expired sources, preventing `GLib-CRITICAL` log spam.

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -122,6 +122,7 @@ const InhibitorManager = GObject.registerClass({
 
         this._isInhibited = false;
         this._inhibitorCookie = null;
+        this._inhibitGeneration = 0;
         this._userEnabled = false;
         this._triggerApp = null;
         this._tempManageLight = false;
@@ -394,23 +395,30 @@ const InhibitorManager = GObject.registerClass({
         }
 
         this._isInhibited = true;
+        const generation = ++this._inhibitGeneration;
         this._sessionManager.InhibitRemote(
             'caffeine-gnome-extension',
             0,
             'Inhibited by Caffeine GNOME extension',
             inhibitFlags,
             (result, error) => {
+                // Ignore stale responses: a remove or a newer inhibit
+                // superseded this request while it was in flight
+                if (generation !== this._inhibitGeneration) {
+                    if (!error) {
+                        const [cookie] = result;
+                        this._sessionManager.UninhibitRemote(cookie);
+                    }
+                    return;
+                }
                 if (error) {
                     this._isInhibited = false;
                     log(`Failed to add inhibitor: ${error}`);
+                    this.emit('update');
                     return;
                 }
                 const [cookie] = result;
-                if (!this._isInhibited) {
-                    this._sessionManager.UninhibitRemote(cookie);
-                } else {
-                    this._inhibitorCookie = cookie;
-                }
+                this._inhibitorCookie = cookie;
             }
         );
     }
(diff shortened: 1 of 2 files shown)


caffeine@patapon.info/extension.js
```diff
@@ -860,16 +872,15 @@ class Caffeine extends QuickSettings.SystemIndicator {
...
-                return GLib.SOURCE_CONTINUE;
-            });
-
-            this._timeOut = GLib.timeout_add(GLib.PRIORITY_DEFAULT, timerDelay * 1000, () => {
-                // Disable Caffeine when timer ended
-                this._removeTimer();
-                if (this._state) {
-                    this._handleToggleClick();
+                if (secondLeft <= 0) {
+                    this._timePrint = null;
```
stuarthayhurst, line 876: Does this really need to be set to `null` if `this._removeTimer()` is called immediately after?

stuarthayhurst:
> Replaced synchronous in extension.js with async InhibitRemote(...)

You mentioned this was for performance, was there an issue with performance? We swapped to synchronous calls because Caffeine kept losing track of the inhibitor, and just calling it synchronously was by far the easiest and most reliable fix.

caffeine@patapon.info/extension.js
ducky-duke: Yes, it's deliberate. We're inside the `_timePrint` source's own callback and about to return `GLib.SOURCE_REMOVE`, so GLib destroys the source for us. Nulling `this._timePrint` first makes `_removeTimer()`'s `if (this._timePrint !== null)` guard skip the `GLib.Source.remove()` call — otherwise we'd remove a source that's currently being dispatched and already about to be auto-removed, which is the double-removal that logs `GLib-CRITICAL: Source ID … was not found`. We still call `_removeTimer()` for the rest of its cleanup (label flush, `_timerEnable = false`). I can add a short comment here since the ordering isn't obvious.

ducky-duke:
Fair point on the inhibitor — the sync `Inhibit` call is cheap and infrequent, so making it async isn't really worth reintroducing the tracking issues you hit before.

Would you prefer I drop that change and keep `Inhibit` synchronous, scoping this PR down to just the MPRIS async discovery (which does block Shell startup) and the timer fixes? I can revisit the inhibitor separately later if it's ever worth it with actual profiling.

caffeine@patapon.info/extension.js
stuarthayhurst: Yeah I think a comment would go a long way here, thanks

stuarthayhurst:
> Would you prefer I drop that change and keep `Inhibit` synchronous

Yes, unless we actually hit performance issues, it's quite nice to not have to worry about race conditions here

> MPRIS async discovery (which does block Shell startup)

I haven't actually looked at this at all, I don't use the feature myself so I'll have to spin up a VM to test it properly. If it does fix a bug and it's the best way to do it then it's definitely worth keeping.

> and the timer fixes

The code looks cleaner (+1), but does this explicitly fix any bugs I can test?

PR #434, opened 2026-07-24
Pause wellbeing break reminders while Caffeine is active

### LINKED ISSUE #398: Feature Request: disable wellbeing when caffeine is active
The new GNOME wellbeing settings darkens the screen when it's time for a break.
I believe it would make sense to add a setting (enabled by default) which disables wellbeing when caffeine is enabled. Because if you don't want the screen to dim, you probably don't want to take breaks during that time.

It would be like integrating [this extension](https://github.com/Fingel/gnome-ext-wellbeing-toggle) but without adding an extra button in the quick settings.

IvanAnishchuk:
If you don't want the screen dimming you probably don't want break reminders going off either, so this pauses GNOME's break reminders while Caffeine is on and puts your list back afterwards.

Same Never/Always/For apps options as the night light control, since it's basically the same problem. Only does anything on GNOME 48+ where break reminders exist, no-op on older shells.

Inspired by [gnome-ext-wellbeing-toggle](https://github.com/Fingel/gnome-ext-wellbeing-toggle), but folded into Caffeine's own state instead of adding a separate quick-settings button, and it plays fine next to that extension if you already run it. Its Wellbeing toggle just follows along, off while Caffeine is on and back on after:

| Caffeine off | Caffeine on |
| --- | --- |
| Caffeine off, Wellbeing on | Caffeine on, Wellbeing off |

Closes #398


stuarthayhurst:
Is there a way to do this without writing to the shell's settings keys? The GNOME extensions reviewers aren't keen on modifying shell settings unless absolutely necessary.

IvanAnishchuk:
IDK, I just followed the way it was done gnome-ext-wellbeing-toggle since that definitely works... I'll research if there are other ways.

PR #435, opened 2026-08-31
feat(shell): Add support for GNOME Shell 51

ChrisLauinger77:
uses deprecated scroll-event; it remains functional in GNOME 51 but should eventually move to Clutter.ScrollController.

### DIFF
--- a/caffeine@patapon.info/extension.js
+++ b/caffeine@patapon.info/extension.js
@@ -705,8 +705,19 @@ class Caffeine extends QuickSettings.SystemIndicator {
 
         // Change user state on icon scroll event
         this._indicator.reactive = true;
-        this._indicator.connectObject('scroll-event',
-            (actor, event) => this._handleScrollEvent(event), this);
+        if (ShellVersion >= 51) {
+            const scrollController = new Clutter.ScrollController({
+                flags: Clutter.ScrollControllerFlags.DISCRETE |
+                    Clutter.ScrollControllerFlags.SCROLL_VERTICAL |
+                    Clutter.ScrollControllerFlags.PHYSICAL_DIRECTION
+            });
+            scrollController.connectObject(
+                'scroll', this._handleScrollController.bind(this), this);
+            this._indicator.add_action(scrollController);
+        } else {
+            this._indicator.connectObject('scroll-event',
+                (_actor, event) => this._handleScrollEvent(event), this);
+        }
 
         // Init position and index of indicator icon
         this.indicatorPosition = this._settings.get_int(INDICATOR_POSITION);
@@ -910,6 +921,14 @@ class Caffeine extends QuickSettings.SystemIndicator {
         this._caffeineToggle.subtitle = text;
     }
 
+    _handleScrollController(_controller, _sprite, _source, _dx, dy) {
+        if (dy < 0) {
+            this._handleScrollDirection(Clutter.ScrollDirection.UP);
+        } else if (dy > 0) {
+            this._handleScrollDirection(Clutter.ScrollDirection.DOWN);
+        }
+    }
+
     _handleScrollEvent(event) {
         // Undo natural scrolling inversions (available on GNOME 49+)
         let scrollDirection = event.get_scroll_direction();
@@ -926,6 +945,10 @@ class Caffeine extends QuickSettings.SystemIndicator {
             }
         }
 
+        this._handleScrollDirection(scrollDirection);
+    }
+
+    _handleScrollDirection(scrollDirection) {
         switch (scrollDirection) {
         case Clutter.ScrollDirection.UP:
             if (!this._state) {
@@ -1091,6 +1114,7 @@ export default class CaffeineExtension extends Extension {
     disable() {
         this._caffeineIndicator.destroy();
         this._caffeineIndicator = null;
+        this._settings = null;
 
         // Unregister shortcut
         Main.wm.removeKeybinding(TOGGLE_SHORTCUT);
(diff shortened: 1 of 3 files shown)


ChrisLauinger77:
I take care about the lint stuff later in the evening.
EGO accepts now upload for 51 ...

stuarthayhurst:
Thanks! If you'd like to you're welcome to port to `Clutter.ScrollController`, otherwise I'll get to it within the next few weeks :)

caffeine@patapon.info/metadata.json
```diff
@@ -1,7 +1,7 @@
 {
-  "version": 60,
+  "version": 61,
```
stuarthayhurst: We shouldn't increment this until we're actually releasing the extension

caffeine@patapon.info/extension.js
```diff
@@ -517,7 +517,14 @@ const CaffeineToggle = GObject.registerClass({
...
         this.menu.addMenuItem(new PopupMenu.PopupSeparatorMenuItem());
-        const settingsItem = this.menu.addAction(_('Settings'), () => Me._openPreferences());
+        const settingsItem = this.menu.addAction(_('Settings'), () => {
+            Me.openPreferences();
+            if (ShellVersion >= 51) {
+                QuickSettingsMenu.menu.close({ fadeOnly: true });
+            } else {
+                QuickSettingsMenu.menu.close(PopupAnimation.FADE);
+            }
+        });
```
stuarthayhurst: Is there a reason this didn't stay as `_openPreferences()`?

ChrisLauinger77: Yes it is only used once - hence I use a arrow function now. No need to have _openPreferences

caffeine@patapon.info/metadata.json
ChrisLauinger77: reverted to 60

ChrisLauinger77:
I did also the clutter thing - i test it now in my VM

caffeine@patapon.info/extension.js
stuarthayhurst: In an effort to keep `_init()` smaller I think it should probably stay as `_openPreferences()`, especially as it's larger than before with the new version check

ChrisLauinger77: OK

ChrisLauinger77:
I am done - VM tests worked 
CI has a bug ...
Brazilian po file had bugs, too

stuarthayhurst:
> I am done - VM tests worked CI has a bug ...

Yeah I noticed that, I think it's probably down to the apt package cache, not 100% sure how to clear it

ChrisLauinger77:
- name: Install build dependencies
  run: |
    sudo apt-get update
    sudo apt-get install --yes make gettext gnome-shell

ChrisLauinger77:
anyways - mergeable from my point of view

stuarthayhurst:
Does this handle inverted scrolling (natural scrolling)? That's the reason behind #402

ChrisLauinger77:
> Does this handle inverted scrolling (natural scrolling)? That's the reason behind #402

yes it does

stuarthayhurst:
> CI has a bug ...

Seems it's failing to find some dependencies on the Azure mirror and not installing `gnome-shell`, then the cache action we use ignores the failure and tries to build the extension anyway. Should fix itself I guess...

caffeine@patapon.info/extension.js
```diff
@@ -705,8 +705,25 @@ class Caffeine extends QuickSettings.SystemIndicator {
...
+                    Clutter.ScrollControllerFlags.PHYSICAL_DIRECTION
+            });
+            scrollController.connect('scroll',
+                (_controller, _sprite, _source, _dx, dy) => {
+                    if (dy < 0) {
+                        this._handleScrollDirection(Clutter.ScrollDirection.UP);
+                    } else if (dy > 0) {
+                        this._handleScrollDirection(Clutter.ScrollDirection.DOWN);
+                    }
+                });
```
stuarthayhurst: Can this be put into something similar to `_handleScrollEvent()`, to keep the scrolling logic in the same place?

stuarthayhurst:
Also working with my testing in a VM, I'm happy to approve after my last comment

caffeine@patapon.info/extension.js
ChrisLauinger77: sure

caffeine@patapon.info/extension.js
```diff
@@ -705,8 +705,18 @@ class Caffeine extends QuickSettings.SystemIndicator {
...
         this._indicator.reactive = true;
-        this._indicator.connectObject('scroll-event',
-            (actor, event) => this._handleScrollEvent(event), this);
+        if (ShellVersion >= 51) {
+            const scrollController = new Clutter.ScrollController({
+                flags: Clutter.ScrollControllerFlags.DISCRETE |
+                    Clutter.ScrollControllerFlags.SCROLL_VERTICAL |
+                    Clutter.ScrollControllerFlags.PHYSICAL_DIRECTION
+            });
+            scrollController.connect('scroll', this._handleScrollController.bind(this));
```
stuarthayhurst: Is there a reason this doesn't use `connectObject()`? I don't see any disconnect from this, so we might be leaking signals.

stuarthayhurst (approved):

caffeine@patapon.info/extension.js
ChrisLauinger77: fixed

stuarthayhurst:
The pipeline failures will be fixed by #436, merging this since GNOME 51 is due today

PR #436, opened 2026-08-31
Update actions/checkout to v7, drop apt cache action

stuarthayhurst:
Update the checkout action, but also as an excuse to debug some failures

### DIFF
--- a/.github/workflows/build-extension.yml
+++ b/.github/workflows/build-extension.yml
@@ -9,7 +9,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@v7
     - uses: awalsh128/cache-apt-pkgs-action@latest
       with:
         packages: make gettext gnome-shell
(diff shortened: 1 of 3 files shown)


stuarthayhurst:
@eonpatapon Should we just drop the apt cache action? We don't get a cache hit very often because the workflows aren't run very often, and the package install is fast enough anyway.

In this case it seems to be failing because a package is missing from a mirror, but that seems to be because the cache action has chosen a 'faster' mirror that's got problems.

stuarthayhurst:
The failures seem to be the same issue as https://github.com/awalsh128/cache-apt-pkgs-action/issues/221, but we've got even more missing packages now than we had a few days ago

stuarthayhurst:
Stripped it out as a test, it's only a little slower but does actually work

stuarthayhurst:
Merging, to avoid the pipeline failures on the new MRs. I've got some more changes planned (to automatically run GNOME's new static analysis tool, to catch issues before it's run at release upload time), but those can come later.

PR #442, opened 2026-09-19
Run Shexli on the extension as a workflow

stuarthayhurst:
Uses `stuarthayhurst/shexli-github-action` to run GNOME's new extension static analysis on the built extension.

It's automatically used on uploaded extensions, so running it early should catch issues before release time. `EGO-M-004` is ignored since it gives false positives for upcoming versions of GNOME.

I've been using it successfully on my own extensions and some of the others I maintain, we can also make warnings fail the pipeline too, but there are too many false-positives for this to be useful.


stuarthayhurst:
Here's the output from the first run:
```
  Warning: [EGO-L-005]: owned object references should be released in disable()
   - Message: Owned references that are cleaned up in `disable()` should also be released with `null` or `undefined`.
   - Code: this._appSystem = Shell.AppSystem.get_default() (caffeine@patapon.info.zip:extension.js:153)
   - URL: https://gjs.guide/extensions/review-guidelines/review-guidelines.html#destroy-all-objects
  Warning: [EGO-M-008]: extensions using unlock-dialog must document it in disable() comments
   - Message: Extensions using `unlock-dialog` should document the reason in `disable()` comments.
   - Code:
      disable() {
          this._caffeineIndicator.destroy();
          this._caffeineIndicator = null;
          this._settings = null;
          // Unregister shortcut
          Main.wm.removeKeybinding(TOGGLE_SHORTCUT);
      } (caffeine@patapon.info.zip:extension.js:1114)
   - URL: https://gjs.guide/extensions/review-guidelines/review-guidelines.html#session-modes
  0 error(s) detected
  2 warning(s) detected (non-fatal)

```

